PR #34, opened 2023-01-16
feat: allow extend on union types

sukovanej:
I'd like to be able to use `extend` on unions of structs.

```typescript
pipe(
  S.struct({ b: S.boolean }),
  S.extend(S.union(
    S.struct({ a: S.literal("a") }),
    S.struct({ a: S.literal("b") })
  ))
)
```

I've added an implementation that enables combination of 
- typeLiteral with union of typeLiterals,
- union of typeLiterals with typeLiteral
- union of typeLiterals with another union of typeLiterals

### DIFF
--- a/src/Schema.ts
+++ b/src/Schema.ts
@@ -461,20 +461,35 @@ export const record: <K extends string | symbol, V>(
   value: Schema<V>
 ) => Schema<{ readonly [k in K]: V }> = I.record
 
+const productTypeLiteralArrays = (xs: ReadonlyArray<AST.AST>, ys: ReadonlyArray<AST.AST>) => {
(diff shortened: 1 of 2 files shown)


gcanti:
Nice idea, thank you @sukovanej

Perhaps we could delegate `productTypeLiteralArrays` to do all the checks and simplify the code of `extend`:

```ts
export const extend = <B>(that: Schema<B>) =>
  <A>(self: Schema<A>): Schema<Spread<A & B>> =>
    make(
      productTypeLiteralArrays(
        AST.isUnion(self.ast) ? self.ast.types : [self.ast],
        AST.isUnion(that.ast) ? that.ast.types : [that.ast]
      )
    )
```

What do you think?

sukovanej:
Hey, love that, looks much cleaner now.

Issue #38, opened 2023-01-18
add multipleOf combinator


gcanti:
## 🚀 Feature request

Add `/data/filter.ts#multipleOf` combinator

### Desired Behavior

Related: [`z.number().multipleOf(5); // Evenly divisible by 5`](https://zod.dev/?id=numbers)

### Suggested Solution

Use `filter` API

jessekelly881:
Closed

PR #39, opened 2023-01-18
feat: added filter/multipleOf

jessekelly881:
Add multipleOf for numeric multiplicity. Closes issue #38

### DIFF
--- a/src/data/filter.ts
+++ b/src/data/filter.ts
@@ -57,6 +57,23 @@ export const greaterThanOrEqualTo = <A extends number>(
 /**
  * @since 1.0.0
  */
+export const multipleOf = <A extends number>(
+  divisor: number,
+  annotationOptions?: AnnotationOptions<A>
+) =>
+  (self: Schema<A>): Schema<A> =>
+    pipe(
+      self,
+      I.filter((a): a is A => a % divisor === 0, {
+        description: `a number divisible by ${divisor}`,
+        jsonSchema: { multipleOf: divisor },
+        ...annotationOptions
+      })
+    )
+
+/**
+ * @since 1.0.0
+ */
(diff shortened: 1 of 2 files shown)


jessekelly881:
The json schema specification specifies that the multipleOf parameter should be a positive number with up to 2 decimal places of precision. How do you think we should handle this?

gcanti:
There's a comment in the stackoverflow answer talking about negative numbers and a small change to the `floatSafeRemainder` function, maybe it's a good idea to add tests for negative and fractional numbers as well

jessekelly881:
Ah. Good point! I updated it. How should we handle the json schema issue btw? https://awwright.github.io/json-schema-org-pages/keyword/multipleOf

gcanti:
> How should we handle the json schema issue btw?

What do you mean?

jessekelly881:
According to the json schema spec the value of multipleOf should be a positive(non-zero) number of at most 2 digits. E.g. -1 or 0.001 are both invalid.

gcanti:
IMO if our schema really works even with `0.0001`, it's a feature we can keep, it's up to a potential JSON Schema compiler to point out that some values are not valid for its specific purpose. 

In general, /schema is capable of performing more validations than JSON Schema and it is perfectly acceptable, we don't necessarily have to follow what JSON Schema does, just make sure that a potential JSON Schema compiler has the ability to understand what it needs to do (and potentially refuse to do it for its own reasons)

jessekelly881:
Makes sense. Also, as a side note, I feel this remainder function might be a useful addition to @fp-ts/data/Number instead of inlining it here.

gcanti:
Thanks @jessekelly881 !

PR #40, opened 2023-01-19
feat: added data/Brand

jessekelly881:
Added data/Brand. Accepts either a `Brand` type or a `RefinedConstructors` to establish the brand. In the first case the brand is simply attached at the type level and in the latter case the value is tested against the `RefinedConstructors.refine` function before the brand is attached.

### DIFF
--- a/src/data/Brand.ts
+++ b/src/data/Brand.ts
@@ -1,0 +1,25 @@
+/**
+ * @since 1.0.0
+ */
+import * as B from "@fp-ts/data/Brand"
+import { pipe } from "@fp-ts/data/Function"
+import * as I from "@fp-ts/schema/internal/common"
+import type * as S from "@fp-ts/schema/Schema"
+
+/**
+ * @since 1.0.0
+ */
+export const brand = <C extends B.Brand<string>>(
+  constructor: B.Brand.Constructor<C>,
+  annotationOptions?: S.AnnotationOptions<B.Brand.Unbranded<C>>
+) =>
+  <A extends B.Brand.Unbranded<C>>(self: S.Schema<A>): S.Schema<A & C> =>
+    B.isNominal(constructor) ?
+      self as S.Schema<A & C> :
+      pipe(
+        self,
+        I.filter<A, A & C>(
+          (x): x is A & C => constructor.refine(x),
+          annotationOptions
+        )
+      )
--- a/test/data/Brand.ts
+++ b/test/data/Brand.ts
@@ -1,0 +1,41 @@
+import * as B from "@fp-ts/data/Brand"
+import { pipe } from "@fp-ts/data/Function"
+import * as _ from "@fp-ts/schema/data/Brand"
+import * as S from "@fp-ts/schema/Schema"
+import * as Util from "@fp-ts/schema/test/util"
+
+type Int = number & B.Brand<"Int">
+const Int = B.refined<Int>(
+  (n) => Number.isInteger(n),
+  (n) => B.error(`Expected ${n} to be an integer`)
+)
+
+type Positive = number & B.Brand<"Positive">
+const Positive = B.refined<Positive>(
+  (n) => n > 0,
+  (n) => B.error(`Expected ${n} to be positive`)
+)
+
+type PositiveInt = Positive & Int
+const PositiveInt = B.all(Int, Positive)
+
+type Eur = number & B.Brand<"Eur">
+const Eur = B.nominal<Eur>()
+
+describe.concurrent("Brand", () => {
+  it("property tests", () => {
+    Util.property(_.brand(Int)(S.number)) // refined
+    Util.property(_.brand(Eur)(S.number)) // nominal
+  })
+
+  it("refined", () => {
+    const schema = pipe(S.number, _.brand(Positive), _.brand(Int))
+
+    Util.expectDecodingFailure(schema, -0.5, "Expected refinement, actual -0.5")
+    Util.expectDecodingFailure(schema, -1, "Expected refinement, actual -1")
+    Util.expectDecodingFailure(schema, 0, "Expected refinement, actual 0")
+    Util.expectDecodingSuccess(schema, 1, 1 as PositiveInt)
+    Util.expectDecodingFailure(schema, 1.5, "Expected refinement, actual 1.5")
+    Util.expectDecodingSuccess(schema, 2, 2 as PositiveInt)
+  })
+})
--- a/.changeset/mighty-feet-warn.md
+++ b/.changeset/mighty-feet-warn.md
@@ -1,0 +1,5 @@
+---
+"@fp-ts/schema": patch
+---
+
+Add support for data/Brand


src/data/Brand.ts
```diff
@@ -0,0 +1,39 @@
...
+ * type Positive = number & B.Brand<"Positive">
+ * const Positive = B.refined<Positive>(
+ *   (n) => n > 0,
+ *   (n) => B.error(`Expected ${n} to be positive`)
+ * )
+ *
+ * pipe(S.number, S.greaterThan(0), brand<Positive>()) // S.Schema<Positive>
+ * pipe(S.number, brand(Positive)) // S.Schema<Positive>
+ */
+export const brand = <C extends B.Brand<string>>(
```
IMax153, line 30: Perhaps an unpopular opinion, but I think the `Brand` constructor should be a required parameter, and we should provide the ability to handle either a `Brand.NominalConstructor` or a `Brand.RefinedConstructors`:

```ts
export const brand = <C extends B.Brand<string>>(
  constructor: B.Brand.Constructor<C>,
  annotationOptions?: S.AnnotationOptions<B.Brand.Unbranded<C>>
) =>
  <A extends B.Brand.Unbranded<C>>(self: S.Schema<A>): S.Schema<A & C> =>
    B.isNominal(constructor) ?
      self as S.Schema<A & C> :
      pipe(
        self,
        I.filter<A, A & C>(
          (x): x is A & C => constructor.refine(x),
          annotationOptions
        )
      )
```

In this way, we get type inference for free:

```ts
import * as B from "@fp-ts/data/Brand";
import * as S from "@fp-ts/schema/Schema";
import { pipe } from "@fp-ts/data/Function";
import { brand } from "@fp-ts/schema/data/Brand";

type Nominal = string & B.Brand<"Nominal">
const Nominal = B.nominal<Nominal>()

pipe(S.string, brand(Nominal)) // S.Schema<Nominal>

type Positive = number & B.Brand<"Positive">
const Positive = B.refined<Positive>(
  (n) => n > 0,
  (n) => B.error(`Expected ${n} to be positive`)
)

pipe(S.number, S.greaterThan(0), brand(Positive)) // S.Schema<Positive>
```

docs/examples/index.ts
```diff
@@ -0,0 +1,2 @@
+import './src-data-Brand.ts-function-brand-0.ts'
```
IMax153, line 1: I think this file may have been left over from docs generation.

docs/examples/src-data-Brand.ts-function-brand-0.ts
```diff
@@ -0,0 +1,13 @@
+import * as B from "@fp-ts/data/Brand";
```
IMax153, line 1: I think this file may have been left over from docs generation.

src/data/Brand.ts
mikearnaldi: ideally if we assume `brand` to be the last combinator the `brand` itself could return a `SchemaWithConctructor` so you could do:

```ts
import * as B from "@fp-ts/data/Brand";
import * as S from "@fp-ts/schema/Schema";
import { pipe } from "@fp-ts/data/Function";
import { brand } from "@fp-ts/schema/data/Brand";

type Nominal = string & B.Brand<"Nominal">
// S.SchemaWithConstructor<Nominal>
const Nominal = pipe(S.string, nominal<Nominal>())

Nominal.of()
Nominal.either()
Nominal.option()
```

jessekelly881: Good point. I had overlooked nominal brands.

test/data/Brand.ts
```diff
@@ -0,0 +1,41 @@
...
+const Eur = B.nominal<Eur>()
+
+describe.concurrent("Brand", () => {
+  it("property tests", () => {
+    Util.property(_.brand(Int)(S.number)) // refined
+    Util.property(_.brand(Eur)(S.number)) // nominal
+  })
+
+  it("refined", () => {
+    const schema = pipe(S.number, _.brand(Positive), _.brand(Int))
```
IMax153, line 32: This is a nitpick, but you could shorten this to:

```ts
const schema = pipe(S.number, _.brand(B.all(Int, Positive)))
```

IMax153 (commented):
I do think we should consider @mikearnaldi's proposal. If we instead make the assumption that `Schema.brand` will always be the final call (not an unreasonable assumption), we could instead return a `SchemaWithConstructor`, which essentially would be a `Schema` for the branded type, with constructors added to it. 

@mikearnaldi - would the idea be that the `Schema` is used to generate the constructors?

jessekelly881:
Yeah, I think that's a fair assumption. How would the SchemaWithConstructor type compose with other schemas though, for example, `S.struct({ age: B.brand(Int) })`? Also, what's the benefit of using this vs just parsing it and folding the value? (e.g. `pipe(-1, P.decode(schema), E.match(() => "Error", () => "Success")`)

test/data/Brand.ts
jessekelly881: Yeah, I actually defined a PositiveInt type in the test doc. The reason for doing it this way was to test/demonstrate composition of brands in the schema.

mikearnaldi:
> I do think we should consider @mikearnaldi's proposal. If we instead make the assumption that `Schema.brand` will always be the final call (not an unreasonable assumption), we could instead return a `SchemaWithConstructor`, which essentially would be a `Schema` for the branded type, with constructors added to it.
> 
> @mikearnaldi - would the idea be that the `Schema` is used to generate the constructors?

Yes at least for branded, we may have similar combinators for example to create a `case` constructor from a schema

mikearnaldi:
> Hmm. I only partly agree with that assumption. Any composition of Brands can be done directly with the Brands and not in the schema, but what about Brands in structs or arrays? For example, `S.struct({ age: B.brand(Int) }) // Struct<{ age: Int }>` or `pipe(S.number, B.brand(Int), S.array) // Struct<Int[]>`? Also, what would be the benefit of using a SchemaWithConstructor vs just parsing the schema? (e.g. `pipe(-1, P.decode(B.brand(Int)), E.match(() => "Error", () => "Success" /* Int */ )`)? I would think that the .of, and .option functions would make more sense on the parser (or in the @fp-ts/data/Brand library) than in the schema. Sorry if I'm missing something obvious here. jaja

They could be extracted out, usually you first create `Int` and then `Int[]` just like you would create the types

jessekelly881:
How do you mean, exactly? Like creating an IntArray brand?

IMax153:
@gcanti @mikearnaldi - we should probably make a decision so that we can get this merged in. 

I think the decision point is:

1. Have `Schema.brand` return a `SchemaWithConstructor`, which would have the pro of providing constructors and a `Schema` at the same time but the con of breaking the ability to continue "pipe-ing" the schema to transform it
2. Delegate constructors elsewhere and have `Schema.brand` simply return the branded schema (as is with the current implementation)

I think after having though about it, my vote would be for option 2.

mikearnaldi:
> @gcanti @mikearnaldi - we should probably make a decision so that we can get this merged in.
> 
> I think the decision point is:
> 
> 1. Have `Schema.brand` return a `SchemaWithConstructor`, which would have the pro of providing constructors and a `Schema` at the same time but the con of breaking the ability to continue "pipe-ing" the schema to transform it
> 2. Delegate constructors elsewhere and have `Schema.brand` simply return the branded schema (as is with the current implementation)
> 
> I think after having though about it, my vote would be for option 2.

A `SchemaWithConstructor` is still a schema that can be composed just fine, it doesn't hurt and it is very common to brand as the last step of a block of composition, that said it doesn't restrict the ability to also directly use constructors from elsewhere, I was planning to add the same construct SchemaWithConstructor to `/schema/data/Data`

jessekelly881:
How should the SchemaWithConstructor work with filters and multiple brands? For example pipe(S.number, S.filter(n => n > 0), S.brand(Positive)). Would the schema with constructor also validate using the filter? Also, what about two consecutive calls to S.brand, for example pipe(S.brand(Positive), S.brand(Int)) would it merge the validators for both brands?

mikearnaldi:
Brand constructors can be composed via brand but it could be made to do
that too. Yes the basic constructor would perform all the checks and throw
in case it isn't valid

jessekelly881:
What would the SchemaWithConstructor type do that S.decode(schema) and S.is(schema) don't do already while being more general(besides the option case which could be added)? Those should work for schemas using S.brand as well.

mikearnaldi:
> What would the SchemaWithConstructor type do that S.decode(schema) and S.is(schema) don't do already while being more general(besides the option case which could be added)? Those should work for "branded schemas" as well.

Neither `S.decode(schema)` nor ` S.is(schema) ` are constructors. A constructor for a branded type will have signature: `(a: Unbranded<A>) => A`

mikearnaldi:
We could even make it such that Schema by default has an `.of` function with signature `(a: A) => A` (basically a validating identity) and that function can be refined by custom schemas

jessekelly881:
Ok. Yeah, I understand now. I was just going to point out that every schema could have a constructor. S.number.of(4) S.tuple([S.number, S.string]).of(""), etc.

mikearnaldi:
> Ok. Yeah, I understand now. I was just going to point out that every schema could have a constructor. S.number.of(4) S.tuple([S.number, S.string]).of([1,""]), etc. So it might makes more sense to add an .of to the Schema type instead of having a separate SchemaWithConstructor type. But this would require updating every schema function, S.number, S.string, S.filter, etc. Also, the signature should probably be

There is a difference between the constructor for a generic schema or for a branded type, or really anything that you'd want custom, probably we should have `SchemaOf<A>`, `SchemaBranded<A>`, `SchemaData<A>` each with it's own flavour of constructor.

https://github.com/fp-ts/schema/pull/73 for example

jessekelly881:
It seems like the SchemaBranded would need two type arguments to correctly implement of, an input type and the branded type. I.e. of: (x: number) => number & Brand<"Positive">.

mikearnaldi:
Not necessarily because number can be derived by un branding A

jessekelly881:
Ah good point! The issue with of still applies though. The SchemaBranded type would presumably only implement the option and either constructors. Unless we want to allow of to throw on failure.

mikearnaldi:
I think we have to allow for throwing on the normal constructors too as all
validation cannot be captured by the types, for example as you mentioned if
you use a filter

jessekelly881:
I think the option and either constructors are probably safe in general across all schemas types because the filters could take in the Schema.option and Schema.either types and chain in the filter. But, yeah, the of constructor is a bit complicated.

jessekelly881:
I think that this is something that needs to be solved across the repo as a while and is probably out of the scope of this PR. What do you think if we merge the current implementation and then create a separate PR for adding constructors across the board. Assuming you are ok with the current implementation of brand pending adding constructors.

jessekelly881:
I think that this is something that needs to be solved across the repo as a while and is probably out of the scope of this PR. What do you think if we merge the current implementation and then create a separate PR for adding constructors across the board. Assuming you are ok with the current implementation of brand pending adding constructors.

mikearnaldi:
I would agree to keep it separate for now

test/data/Brand.ts
jessekelly881: Resolving as is unless you think it should be changed.

mikearnaldi (approved):

PR #42, opened 2023-01-19
feat: add DateKeyword to AST

TylorS:
Adds a `DateKeyword` to the AST and implements it for all of the core compilers and the test compilers

### DIFF
--- a/src/AST.ts
+++ b/src/AST.ts
@@ -33,6 +33,7 @@ export type AST =
   | BigIntKeyword
   | SymbolKeyword
   | ObjectKeyword
+  | DateKeyword
   | Enums
   | TemplateLiteral
   | Tuple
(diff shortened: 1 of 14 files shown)


src/AST.ts
```diff
@@ -33,6 +33,7 @@ export type AST =
   | BigIntKeyword
   | SymbolKeyword
   | ObjectKeyword
+  | DateKeyword
```
photz, line 36: My understanding is that the AST type should mirror Typescripts' actual AST, and there is no keyword or separate syntax for calendar dates. 
Perhaps it would be better to implement support for dates in userland?

gcanti:
Closing in favour of https://github.com/fp-ts/schema/pull/43

PR #43, opened 2023-01-19
feat: add date

tim-smart:
Do we want to make parseDate more strict?


gcanti:
I think it's good as it is, if someone wants to be more strict they can always pass a more fitting Schema<string> as input to `parseDate`

PR #45, opened 2023-01-22
fix: extend union with mutually exclusive properties

sukovanej:
Hey,

after [adding the support for union types](https://github.com/fp-ts/schema/pull/34), the `extend` allows the following construction

```typescript
pipe(
  S.struct({ a: S.literal("a") }),
  S.extend(
    S.union(
      S.struct({ a: S.literal("b"), b: S.string }),
      S.struct({ a: S.literal("a"), b: S.number })
    )
  )
)
```

which has an expected type

```typescript
S.Schema<{
  readonly a: "a";
  readonly b: number;
}>
```

but in the term-level code the AST is a union where each element has a duplicated `a` literal. As a consequence, fast-check generates a non-sensical object `{"a":"b","b":""}`.

One solution is simply not supporting overlapping properties for the `extend` combinator but I find it super useful to be able to extend with an union where each element can discriminate with its field.

Therefore, in this MR I'm proposing to move the implentation of `extend` closer to what `Spread` does in the type-level code. I added more general `intersectTypeLiterals` function which returns `Option<AST.TypeLiteral>` to be able to skip mutually exclusive combinations of `TypeLiteral`s. It uses `areMutuallyExclusive` function which currently supports only `Literal`s but I can imagine it could be implemented for other AST variants. 

*Side-note: I'm currently in a process of an experimental rewrite from `io-ts` in our project and this is the last bit I was missing to have a successful prototype. That's why I focused on the `TypeLiteral` only.*

A similar issue happens for the following schema

```typescript
pipe(
  S.struct({ a: S.struct({ b: S.string }) }),
  S.extend(
    S.struct({ a: S.struct({ b: S.string, c: S.number }) }),
  )
)

// --> type
S.Schema<{
  readonly a: {
    readonly b: string;
  } & {
    readonly b: string;
    readonly c: number;
  };
}>
```

for which `P.is(schema)` works incorrectly / unexpectedly. The `areMutuallyExclusive` can catch it and at least it fails immediately which is probably better than having unexpected behaviour on derived artifacts. I suppose, it will be necessary to implement a general AST intersection that behaves the same way as the type intersection in the future.

@gcanti, please let me know what you think. 

And thank you for your awesome work on this and all the other projects. `@fp-ts/schema` feels like it's gonna be a game changer!

### DIFF
--- a/src/Schema.ts
+++ b/src/Schema.ts
@@ -5,6 +5,7 @@
 import { pipe } from "@fp-ts/data/Function"
 import type { Json } from "@fp-ts/data/Json"
 import type { Option } from "@fp-ts/data/Option"
+import * as O from "@fp-ts/data/Option"
 import type { Predicate, Refinement } from "@fp-ts/data/Predicate"
 import * as RA from "@fp-ts/data/ReadonlyArray"
 import * as A from "@fp-ts/schema/annotation/AST"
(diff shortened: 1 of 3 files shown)


gcanti:
Great, well done. I also think that we need a more powerful `intersect` function that works with more generic ASTs (although I think that some limitations are needed to not overcomplicate the code).

gcanti:
> One solution is simply not supporting overlapping properties for the extend combinator

@sukovanej I've thought about it a lot today and I've come to the conclusion that an API that is called `extend` should indeed only allow extensions, not overwrites. It seems confusing to open the possibility of overlapping properties but then only handle a particular case.

Furthermore, if we allow overlapping properties, there are cases where the result might be surprising

```ts
const schema = pipe(
  S.struct({ a: S.literal("a") }),
  S.extend(
    S.struct({ a: pipe(S.literal("a"), S.identifier("a")) })
  )
)
// this fails
expect(schema).toStrictEqual(S.struct({ a: pipe(S.literal("a"), S.identifier("a")) }))
```

I would revert the change, and even add stricter checks to make sure that properties or index signatures that are already present cannot be overwritten.

PR #53, opened 2023-01-25
Introduce Match

mikearnaldi:


jessekelly881:
Using a _with_ and a _not_ with a literal value does not seem to be marked as exhaustive. For example,

```js
const match = pipe(
  Match.type<number>(),
  Match.when(1, (_) => "a"),
  Match.not(1, (_) => "a"),
  Match.exaustive // Type error
)
```

mikearnaldi:
> Using a _with_ and a _not_ with a literal value does not seem to be marked as exhaustive. For example,
> 
> ```js
> const match = pipe(
>   Match.type<number>(),
>   Match.when(1, (_) => "a"),
>   Match.not(1, (_) => "a"),
>   Match.exaustive // Type error
> )
> ```
> 
> Otherwise, this PR looks awesome! I'm super excited for this to be merged.

Not sure it will end up in schema probably if this gets through it will be it's own library I've just placed it here while we were exploring given it depends on schema (similar strategy to the Arbitrary module that will be extracted out)

tim-smart:
> Using a _with_ and a _not_ with a literal value does not seem to be marked as exhaustive. For example,
> 
> ```js
> const match = pipe(
>   Match.type<number>(),
>   Match.when(1, (_) => "a"),
>   Match.not(1, (_) => "a"),
>   Match.exaustive // Type error
> )
> ```
> 
> Otherwise, this PR looks awesome! I'm super excited for this to be merged.

Hmm will need to think about that one some more, as `Exclude<number, 1>` reduces to `number`.

tim-smart:
> Using a _with_ and a _not_ with a literal value does not seem to be marked as exhaustive. For example,
> 
> ```js
> const match = pipe(
>   Match.type<number>(),
>   Match.when(1, (_) => "a"),
>   Match.not(1, (_) => "a"),
>   Match.exaustive // Type error
> )
> ```
> 
> Otherwise, this PR looks awesome! I'm super excited for this to be merged.

Resolved here: https://github.com/fp-ts/schema/pull/60/files

jessekelly881:
> > Using a _with_ and a _not_ with a literal value does not seem to be marked as exhaustive. For example,
> > ```js
> > const match = pipe(
> >   Match.type<number>(),
> >   Match.when(1, (_) => "a"),
> >   Match.not(1, (_) => "a"),
> >   Match.exaustive // Type error
> > )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Otherwise, this PR looks awesome! I'm super excited for this to be merged.
> 
> Resolved here: https://github.com/fp-ts/schema/pull/60/files

Oh cool! Is it possible to add add predicates as well? https://github.com/gvergnaud/ts-pattern#Pwhen-patterns

tim-smart:
@jessekelly881 I'll look at adding predicates soon.

Will need to work out the inference a bit more.

jessekelly881:
Why the choice to move this to the effect org  instead of the @fp-ts org out of curiosity? This seems deeply integrated with fp-ts and not very specific to effect.

mikearnaldi:
It will need to depend on types from both fp-ts and effect for the unify functionality and given also /data is getting moved to effect it kind of make sense having it here and fp-ts wants a closed scope core/optic/schema

PR #54, opened 2023-01-25
Schema/match changes for Mike

tim-smart:


test/Match.ts
```diff
@@ -38,13 +38,35 @@ describe("Match", () => {
...
   })
 
+  it("tuples", () => {
+    const match = pipe(
+      Match.type<[string, string]>(),
+      Match.when(["yeah"], (_) => true),
+      Match.option
+    )
+
+    expect(match(["yeah", "a"])).toEqual({ _tag: "Some", value: true })
```
datner, line 48: might be nicer to just:
```ts
assert.isStrictEqual(match(["yeah", "a"]), O.some(true))
```
also more resistant to interface changes

datner: (like line 48)

Issue #48, opened 2023-01-23
Parser should support typed Input "I" and Output "O"


liambutler-lawrence:
## 🐛 Bug report

I'm porting over our code from io-ts (big improvement!) but am running into issues with `P.encode`.

### Current Behavior

`P.encode(S.literal('foo'))(<some string>)` returns `unknown` instead of `string`. We have a lot of branded (now filtered) and other custom types that we use in our codebase, but we also have lower-level code (e.g. persistence layers, math-heavy algos) that operate over primitives like `io-ts-types`'s `Json` type. We bridge between those two realms with `encode`, which now loses type information and causes lots of compiler errors.

### Expected behavior

In `io-ts`, `t.literal('foo').encode(<some string>)` returned a `string`, not `unknown`.

### Reproducible example

See above.

### Suggested solution(s)

Support primitive input/output types by default (see below as well).

### Additional context

`io-ts` already had this problem with `decode` as well, where the built-in types (e.g. `t.string`) defaulted the `I` type to `unknown`. But the consequence of that was only reduced type safety when bridging from primitives to complex types, whereas the consequence going the other direction (with `encode`) is actual compiler errors, requiring lots of unsafe casts to get our existing code working.

### Your environment

| Software      | Version(s) |
| ------------- | ---------- |
| @fp-ts/schema |    0.0.8        |
| TypeScript    |    4.8.4        |

liambutler-lawrence:
Potential workarounds for the short term are also welcome, I'm aware schema is still in beta :)

gcanti:
In `io-ts`, the main abstraction keeps track of three types, but in `@fp-ts/schema` we've made the decision to only keep track of one type as a design choice (for the time being the main goal is to represent a TypeScript type at runtime).

`Parser` is a artifact produced by a `Schema` and since `Schema` has only one type parameter it is not possible for its derivatives to have more.

I think it is possible to define a different abstraction from `Schema` that still utilizes the `AST` module under the hood and keeps track of additional types depending on the specific use case, but it is something that would need to be studied from the ground up.

jessekelly881:
I think the issue is just an incorrect type casting. This PR(#56) should fix it.

stazz:
Looks like this changed in https://github.com/Effect-TS/schema/commit/df9ea093623640a0e06cdd0147e8e7522ee05110 ?

gcanti:
yeah, people have been wanting a second type parameter for quite some time, there have been several discussions about it in effect Discord

PR #56, opened 2023-01-25
DRAFT: proposed change to the ast implementation

jessekelly881:
Fixed type casting. Fixes issue #48 

The _Parser.parse_ function returns a `ParseResult<A>` but it is being cast to `ParseResult<unknown>` in the `encoder` function. The relevant code is:

```js
export interface Parser<A> extends Schema<A> {
  readonly parse: (input: unknown, options?: ParseOptions) => ParseResult<A>
}

export const encode = <A>(
  schema: Schema<A>
): (a: A, options?: ParseOptions) => ParseResult<unknown /* should be A */> => parserFor(schema, "encoder").parse
```

### DIFF
--- a/src/Parser.ts
+++ b/src/Parser.ts
@@ -86,7 +86,7 @@ export const asserts = <A>(schema: Schema<A>) =>
  */
 export const encode = <A>(
   schema: Schema<A>
-): (a: A, options?: ParseOptions) => ParseResult<unknown> => parserFor(schema, "encoder").parse
+): (a: A, options?: ParseOptions) => ParseResult<A> => parserFor(schema, "encoder").parse
 
 /**
  * @category encoding
--- a/src/Parser.ts
+++ b/src/Parser.ts
@@ -86,7 +86,7 @@ export const asserts = <A>(schema: Schema<A>) =>
  */
 export const encode = <A>(
   schema: Schema<A>
-): (a: A, options?: ParseOptions) => ParseResult<A> => parserFor(schema, "encoder").parse
+): ((a: A, options?: ParseOptions) => ParseResult<A>) => parserFor(schema, "encoder").parse
 
 /**
  * @category encoding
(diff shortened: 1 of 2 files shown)


gcanti:
> ParseResult<unknown /* should be A */>

In general it's not true

```ts
import { pipe } from "@fp-ts/core/Function"
import * as S from "@fp-ts/schema"
import { parseNumber } from "@fp-ts/schema/data/parser"

expect(S.encodeOrThrow(pipe(S.string, parseNumber))(1)).toEqual("1")
```

jessekelly881:
Ah. I see. The issue is a bit deeper than just a casting issue. The parserFor function isn't actually doing the type transformation, it's just returning Parcer<A> for everything. Have you considered expanding the Schema type so that it can better keep track of the underlying AST? I've played with this for adding proper type checking to the rest function.  

```js
export interface Schema<A, B extends AST.AST> {
  readonly A: (_: A) => A
  readonly ast: B
}

const number: Schema<number, AST.NumberKeyword> = I.number
// etc. 
```

gcanti:
> Have you considered expanding the Schema type so that it can better keep track of the underlying AST?

Only one type parameter is needed since a `Schema<A>` is the runtime representation of the TypeScript type `A`

jessekelly881:
@gcanti You could also derive the _A from the AST. I added a test implementation to this pr. There are actually a lot of benefits to doing it this way. I've implemented full function schemas, intersection types, maps between types(preserving type). You could also implement optics between schemas using this implementation. It should also preserve types better between function calls. The only downside that I can see vs the current implementation is with typeLiterals but I think that they are possible with this implementation.

gcanti:
I'm not sure I'm following, `encode` cannot have the following signature

```ts
export const encode = <A>(
  schema: Schema<A>
) => (a: A, options?: ParseOptions): ParseResult<A> => ...
```

in any implementation that takes into account a parsing process and not just validation.

In addition, if the implementation only performs validation, then an `encode` function is not even necessary.

By parsing, I mean a transformation like the one I showed above, from string to number, for example.
By validation, I mean just the assertion of a type, for example that an input is a string, and nothing else.

PR #57, opened 2023-01-25
fix: encoder refinement

sukovanej:
Hey, during implementation of a schema for an integer parsed from a string (running a refinement on top of a transformation) I noticed the encoder returns a failure unexpectedly. Here's my example schema.

```typescript
pipe(P.parseNumber(S.string), S.int())
```

The problem seems to be in the `Refinement` implementation. It runs the refinement check after running parsing on the `ast.from`. I think for the `encoder` case it should be the other way around and it looks like it fixed the issue.

### DIFF
--- a/src/Parser.ts
+++ b/src/Parser.ts
@@ -426,14 +426,22 @@ const parserFor = <A>(
       }
       case "Refinement": {
         const type = go(ast.from)
-        return make(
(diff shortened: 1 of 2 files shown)


gcanti:
Good catch, thank you @sukovanej

PR #61, opened 2023-01-25
fix: type literal signatures priority

sukovanej:
I encoutered a problem when using `extend` with `record`. Actually an example of it is currently in the readme.

```typescript
// $ExpectType Schema<{ [x: string]: string; readonly a: string; readonly b: string; readonly c: boolean; }>
pipe(
  S.struct({ a: S.string, b: S.string }),
  S.extend(S.struct({ c: S.boolean })), // <= you can add more fields
  S.extend(S.record(S.string, S.string)) // <= you can add index signatures
);
```

The problem is when index and property signatures are mutually exclusive (in the example `boolean` and `string`). The typescript simply prefers property signatures over index signature thus I believe keys covered by property signatures should be skipped by index signatures. WDYT?


sukovanej:
Changed to a draft. Found issue for `Pretty` .

```typescript
  it("extend with record", () => {
    const pretty = P.pretty(
      pipe(S.struct({ a: S.string }), S.extend(S.record(S.string, S.string)))
    );
    expect(pretty({ a: "a" })).toEqual(`{ "a": "a" }`);
  });
```

Fails on 

```
  - Expected   "{ "a": "a" }"
  + Received   "{ "a": "a", "a": "a" }"
```

gcanti:
> The typescript simply prefers property signatures over index signature

It throws an error:

```ts
type A = {
  [x: string]: boolean
  readonly a: string // <= error here
}
```

> Property 'a' of type 'string' is not assignable to 'string' index type 'boolean'.ts(2411)

Ideally, we shouldn't allow it either in /schema

gcanti:
I'm not happy about cluttering the code with all these checks (and it doesn't seem easy to do them well either, so I'm doubly unhappy), unfortunately it seems that TypeScript allows for these type-level inconsistencies when using intersections, even though it raises an error on value-level

```ts
export type A = {
  [x: string]: boolean
  readonly a: string // error: roperty 'a' of type 'string' is not assignable to 'string' index type 'boolean'.ts(2411)
}

export const a: A = { a: "a" } // Property 'a' is incompatible with index signature

export type B = {
  readonly b: string // no error
} & {
  readonly [x: string]: boolean
}

export const b: B = { b: "b" } // Property 'b' is incompatible with index signature
```

sukovanej:
Ah, you're right, I tried it only with the intersection and I didn't even notice the problem on assignment. Good catch.

With these findings, I think it simply shouldn't be allowed. The only non-faulty combination seems to be a case when index and property signatures have the same value type.

I'm going to close this MR but I'd like to discuss these two things.

1. `Pretty` in the current state won't work properly for any combination of property and index signatures, it will duplicate keys in that case. There, I'm afraid, keys touched by property signatures must be skipped.
2. I was working on this issue because I need an extend with an unknown record to allow for unknown attributes inside of nested structures. Would you have any idea how to enable this without the need to implement a custom extend in the userland?

gcanti:
> The only non-faulty combination seems to be a case when index and property signatures have the same value type

I think that the constraint is more relaxed, i.e. the property signature must be a subtype (so in your case, `unknown`, there should be no problems)

> I'm going to close this MR

I think this PR is okay.

Pretty is definitely bugged and needs to be fixed.

As for Arbitrary and Parser, I am currently favoring index signatures over property signatures, but the opposite, which is what this PR implements, seems more appropriate and less surprising for users.

src/Parser.ts
```diff
@@ -325,6 +325,7 @@ const parserFor = <A>(
                 output[name] = t.right
               }
             }
+            const resolvedKeys = Object.keys(output)
```
gcanti, line 328: It seems to me that we are already tracking the processed keys in the `expectedKeys` map, it can be reused

gcanti (changes requested):

src/Parser.ts
```diff
@@ -334,6 +335,10 @@ const parserFor = <A>(
                 const type = indexSignatures[i][1]
                 const keys = I.getKeysForIndexSignature(input, ast.indexSignatures[i].parameter)
                 for (const key of keys) {
+                  if (resolvedKeys.includes(key as string)) {
```
gcanti, line 338: Here, if we use `if (Object.prototype.hasOwnProperty.call(expectedKeys, key)) {` we can avoid the cast.

src/Pretty.ts
```diff
@@ -148,6 +150,9 @@ const prettyFor = <A>(schema: Schema<A>): Pretty<A> => {
                 const type = indexSignatureTypes[i]
                 const keys = I.getKeysForIndexSignature(input, ast.indexSignatures[i].parameter)
                 for (const key of keys) {
+                  if (resolvedKeys.includes(key as string)) {
```
gcanti, line 153: Here, I would standardize the way of tracking processed keys to the method used in `Parser`

sukovanej:
> I think that the constraint is more relaxed (*), i.e. the property signature must be a subtype (so in your case, unknown, there should be no problems)

Yes, I'd expect exactly that. It seems to be okay with union types but for example with objects, I noticed the same issue with a literal assignment.

```typescript
type S = { a: string };

type A = {
  readonly a: S & { b: string };
  readonly [x: string]: S;
};

const a: A = { a: { a: 'string', b: 'string' } };

/**
Type '{ a: { a: string; b: string; }; }' is not assignable to type 'A'.
  Property 'a' is incompatible with index signature.
    Type '{ a: string; b: string; }' is not assignable to type 'S'.
      Object literal may only specify known properties, and 'b' does not exist in type 'S'.
*/
```

It seems to be similar to the one with boolean and number. For both, the typescript complains during assignment but once I have a value of type `A`, typescript treats types "correctly".

```
type A = {
  readonly a: boolean;
} & {
  readonly [x: string]: string;
};

const a: A = { a: 1 }; // Type 'number' is not assignable to type 'boolean'.
type X = typeof a.a // => boolean
type Y = typeof a.whatever // => string
```

gcanti:
Looks like "Excess property checking" in action:

```ts
type S = { a: string }

type A = {
  readonly a: S & { b: string }
  readonly [x: string]: S
}

export const a: A = { a: { a: "string", b: "string" } } // error

const b = { a: { a: "string", b: "string" } }

export const a2: A = b // no error
```

src/Parser.ts
sukovanej: 👍  applied

sukovanej:
> Looks like "Excess property checking" in action
Makes sense 👍 althought it looks weird to me.

One more thing I'm unsure about.

```typescript
  it("extend/record/ record(string, NumberFromString)", () => {
    const schema = pipe(
      S.struct({ a: S.number }),
      S.extend(S.record(S.string, NumberFromString))
    )
    Util.expectEncodingSuccess(schema, { a: 1 }, { a: 1 })
    Util.expectEncodingSuccess(schema, { a: 1, b: 1 }, { a: 1, b: "1" })
  })
```

If the idea is that index signature should cover non-property signatures only, the behaviour makes sense. But if I thought about combinations of index a property signatures in a way that index signature must apply for the whole object and on top of it, property signatures must apply, then I would rather expect the original behaviour. What's your take on that, please?

gcanti:
We have to make a choice, even though this choice may prove to be arbitrary, given that in the case there are transformations there may inevitably be conflicts, both in the decoding phase and in the encoding phase.

In case of conflicts we have 4 alternatives:
1) we favor index signatures over property signatures (as it is now)
2) we favor property signatures over index signatures (as in this PR)
3) we find a way to resolve conflicts that always works
4) we find a way to let users choose what to do in case of conflicts

I think that between 1) and 2), the options we have ready at this moment, it is better to move on to 2)

gcanti (approved):

gcanti:
Thank you @sukovanej , released

PR #76, opened 2023-02-06
refactor: organized filters in data/filters and data/parsers by type

jessekelly881:
Moved filters and parsers from data/filter and data/parser into data/Number, data/String, and data/Object. The interface for "S._" is unchanged but the filters should be moved into files based on their respective type for better organization and to prevent namespace clashes.


IMax153:
I'm not sure if this impacts circularity at all - we should add a circularity check with `madge` similar to what we do in the Effect repos @mikearnaldi

jessekelly881:
Running `madge --ts-config ./tsconfig.madge.json --circular --no-color --no-spinner --warning build/esm` (copied from the @effect-ts/effect repo) showed no circular deps. But yeah, adding this to the github action would be nice.

gcanti (approved):

PR #78, opened 2023-02-08
added data/ReadonlyArray

jessekelly881:
Added data/ReadonlyArray

### DIFF
--- a/src/data/ReadonlyArray.ts
+++ b/src/data/ReadonlyArray.ts
@@ -1,0 +1,79 @@
+/**
+ * @since 1.0.0
+ */
+
+import { pipe } from "@fp-ts/core/Function"
+import * as I from "@fp-ts/schema/internal/common"
+import type { AnnotationOptions, Schema } from "@fp-ts/schema/Schema"
+
+/**
+ * @category identifiers
+ * @since 1.0.0
+ */
+export const MinItemsId = "@fp-ts/schema/ReadonlyArray/minItems"
+
+/**
+ * @category identifiers
+ * @since 1.0.0
+ */
+export const MaxItemsId = "@fp-ts/schema/ReadonlyArray/maxItems"
+
+/**
+ * @category identifiers
+ * @since 1.0.0
+ */
+export const ItemsCountId = "@fp-ts/schema/ReadonlyArray/itemsCount"
+
+/**
+ * @since 1.0.0
+ */
+export const minItems = <A>(
+  n: number,
+  annotationOptions?: AnnotationOptions<ReadonlyArray<A>>
+) =>
+  (self: Schema<ReadonlyArray<A>>): Schema<ReadonlyArray<A>> =>
+    pipe(
+      self,
+      I.filter((a): a is ReadonlyArray<A> => a.length >= n, {
+        description: `an array of at least ${n} items`,
+        jsonSchema: { minItems: n },
+        custom: { id: MinItemsId },
+        ...annotationOptions
+      })
+    )
+
+/**
+ * @since 1.0.0
+ */
+export const maxItems = <A>(
+  n: number,
+  annotationOptions?: AnnotationOptions<ReadonlyArray<A>>
+) =>
+  (self: Schema<ReadonlyArray<A>>): Schema<ReadonlyArray<A>> =>
+    pipe(
+      self,
+      I.filter((a): a is ReadonlyArray<A> => a.length <= n, {
+        description: `an array of at most ${n} items`,
+        jsonSchema: { maxItems: n },
+        custom: { id: MaxItemsId },
+        ...annotationOptions
+      })
+    )
+
+/**
+ * @since 1.0.0
+ */
+export const itemsCount = <A>(
+  n: number,
+  annotationOptions?: AnnotationOptions<ReadonlyArray<A>>
+) =>
+  (self: Schema<ReadonlyArray<A>>): Schema<ReadonlyArray<A>> =>
+    pipe(
+      self,
+      I.filter((a): a is ReadonlyArray<A> => a.length === n, {
+        description: `an array of exactly ${n} items`,
+        jsonSchema: { minItems: n, maxItems: n },
+        custom: { id: ItemsCountId },
+        ...annotationOptions
+      })
+    )
--- a/.changeset/unlucky-ants-learn.md
+++ b/.changeset/unlucky-ants-learn.md
@@ -1,0 +1,5 @@
+---
+"@fp-ts/schema": patch
+---
+
+added data/ReadonlyArray and filtes
--- a/README.md
+++ b/README.md
@@ -484,6 +484,14 @@ pipe(S.number, S.negative()); // < 0
 pipe(S.number, S.nonPositive()); // <= 0
 ```
 
+### Array filters
+
+```ts
+pipe(S.array(S.number), A.maxItems(2)) // max array length
+pipe(S.array(S.number), A.minItems(2)) // min array length
+pipe(S.array(S.number), A.itemsCount(2)) // exact array length
+```
+
 ## Native enums
 
 ```ts
--- a/src/Schema.ts
+++ b/src/Schema.ts
@@ -13,6 +13,7 @@ import * as DataDate from "@fp-ts/schema/data/Date"
 import * as N from "@fp-ts/schema/data/Number"
 import * as O from "@fp-ts/schema/data/Object"
 import * as DataOption from "@fp-ts/schema/data/Option"
+import * as SRA from "@fp-ts/schema/data/ReadonlyArray"
 import * as S from "@fp-ts/schema/data/String"
 import * as I from "@fp-ts/schema/internal/common"
 import type { ParseResult } from "@fp-ts/schema/ParseResult"
@@ -339,6 +340,33 @@ export const nonNegative: <A extends number>(
 export const nonPositive: <A extends number>(
   options?: AnnotationOptions<A>
 ) => (self: Schema<A>) => Schema<A> = N.nonPositive
+
+/**
+ * @category filters
+ * @since 1.0.0
+ */
+export const maxItems: <A>(
+  n: number,
+  options?: AnnotationOptions<ReadonlyArray<A>>
+) => (self: Schema<ReadonlyArray<A>>) => Schema<ReadonlyArray<A>> = SRA.maxItems
+
+/**
+ * @category filters
+ * @since 1.0.0
+ */
+export const minItems: <A>(
+  n: number,
+  options?: AnnotationOptions<ReadonlyArray<A>>
+) => (self: Schema<ReadonlyArray<A>>) => Schema<ReadonlyArray<A>> = SRA.minItems
+
+/**
+ * @category filters
+ * @since 1.0.0
+ */
+export const itemsCount: <A>(
+  n: number,
+  options?: AnnotationOptions<ReadonlyArray<A>>
+) => (self: Schema<ReadonlyArray<A>>) => Schema<ReadonlyArray<A>> = SRA.itemsCount
 
 // ---------------------------------------------
 // combinators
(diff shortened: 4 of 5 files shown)


src/data/ReadonlyArray.ts
```diff
@@ -0,0 +1,75 @@
...
+        description: `an array of at most ${n} elements`,
+        jsonSchema: { maxItems: n },
+        ...annotationOptions
+      })
+    )
+
+/**
+ * @since 1.0.0
+ */
+export const nonEmpty = <A>(
```
gcanti, line 47: Note that in `Schema.ts` there's a `nonEmptyArray` combinator

gcanti:
Perhaps it's better to call them `minItems` and `maxItems` so we can export them from `Schema.ts` without conflicts with `minLength` and `maxLength`.

src/data/ReadonlyArray.ts
jessekelly881: Oops. I didn't see this. I assume that we prefer name "nonEmptyArray" to prevent namespace collisions.

jessekelly881:
Yeah, that's probably a good idea. We could also consider overriding the S.maxLength function to call the appropriate type specific maxLength fn. I.e. given a string it uses the string maxLength fn, for arrays it uses the array variant, for sets the set variant, etc.

gcanti:
> We could also consider overloading the S.maxLength function

Yeah, I have thought about it, but I believe keeping things separate is ultimately less complicated, and after all we are using well-known names, those from JSON Schema.

jessekelly881:
@gcanti I updated it.

jessekelly881:
I feel like we are going to have a lot of issues with name conflicts if we want basically everything to accessible via the top level api (i.e. S._). I suspect that it makes sense to either not have anything in the top level api except core functionality and then have to import say "@fp-ts/schema/data/String" to get string length, minLength, etc. for each object. Or to override the length, minLength, etc. functions to work with every supported object type. Once we have a few things that want the same name (for example, S.length for strings, arrays, sets, etc. it starts to become pretty confusing.

gcanti:
I don't think it's important to export everything, only the most common and frequently used APIs, and now it seems to me that we have reached this point, if we take the zod APIs as a reference, there is almost parity.

src/data/ReadonlyArray.ts
```diff
@@ -0,0 +1,61 @@
...
+export const maxItems = <A>(
+  n: number,
+  annotationOptions?: AnnotationOptions<ReadonlyArray<A>>
+) =>
+  (self: Schema<ReadonlyArray<A>>): Schema<ReadonlyArray<A>> =>
+    pipe(
+      self,
+      I.filter((a): a is ReadonlyArray<A> => a.length <= n, {
+        description: `an array of at most ${n} items`,
+        custom: { type: "maxItems" },
```
gcanti, line 39: Regarding `custom`, it seems that the convention so far has been to put it only if there is no exact counterpart in JSON Schema

src/data/ReadonlyArray.ts
```diff
@@ -0,0 +1,61 @@
...
+        jsonSchema: { maxItems: n },
+        ...annotationOptions
+      })
+    )
+
+/**
+ * @since 1.0.0
+ */
+export const itemsCount = <A>(
```
gcanti, line 48: Regarding `itemsCount`, we would need to have a broader discussion that I haven't yet tackled because in /data/String there's only `minLength` and `maxLength` and `length` is missing: should we define a filter from scratch or implement it through the combination of min* and max*?

jessekelly881: I think having all three makes the most sense intuitively. zod and yup for example have all three(min, max, length) for strings. I think it makes sense to copy the user facing api from both as much as possible to make code migration from one library to another as easy as possible.

gcanti: I'm not talking about the APIs, I am convinced that they are convenient and therefore should be added. I'm talking about the implementation of `itemsCount` and `length` (which is missing by the way). The implementations could be defined from scratch, or make use of existing combinators

src/data/ReadonlyArray.ts
jessekelly881: Oh I see. I think it would be better to have each filter define a "property" on the type. E.g. "positive", "int", etc. and then have a well defined PropertyID type for each filter that exported from the module and is attached to each filter. E.g. 

```js
export type MaxItemsPropertyId = "@fp-ts/schema/data/ReadonlyArray/maxItems"

I.filter((a): a is ReadonlyArray<A> => a.length <= n, {
        description: `an array of at most ${n} items`,
        custom: { type: MaxItemsPropertyId },
})
```

That way it could be uniquely referenced later on and by compilers other than jsonSchema.

src/data/ReadonlyArray.ts
jessekelly881: https://jsbench.me/r2ldwyn11m

jessekelly881: I think performance wise the fewer function calls that are made the better so defining it from scratch would be more performant. At the risk of being a bit more code.

gcanti: In my personal opinion, implementing `itemsCount` and `length` based on `minItems` and `maxItems` (or `minLength` and `maxLength` respectively) is just a complication and not worth it, but I wanted to start the discussion just to give the opportunity to discuss it a bit. If there are no doubts, your current implementation of `itemsCount` is fine for me

src/data/ReadonlyArray.ts
gcanti: I agree, maybe we should change from `type` to `id` then:

```ts
export type MaxItemsId = "@fp-ts/schema/data/ReadonlyArray/maxItems"

I.filter((a): a is ReadonlyArray<A> => a.length <= n, {
        description: `an array of at most ${n} items`,
        custom: { id: MaxItemsId },
})
```

jessekelly881: That would probably be better, yeah. Actually, could we just use the identifier key instead of stuffing it under custom? I.e. 

```
export type MaxItemsId = "@fp-ts/schema/data/ReadonlyArray/maxItems"

I.filter((a): a is ReadonlyArray<A> => a.length <= n, {
        description: `an array of at most ${n} items`,
        identifier: MaxItemsId
})
```

jessekelly881:
@gcanti  I removed the custom tag and updated each of the functions to use an identifier as discussed in the thread.

src/data/ReadonlyArray.ts
jessekelly881: Hmm. Actually, having tried this it seems like identifier is being used for other things like the Typescript compiler. Maybe we could extend the type object to include a "tag" or "key"? Or repurpose the _identifier_ key.

gcanti: Yes, "identifier" is supposed to be used to specify an identifier that can be used safely in code generation, such as by the TypeScript compiler in /test, or by other compilers that do code generation.

"custom" is a provisional name that I put in but the more I think about it the less I like it, since it's only used for filters for now we could even rename it to "filter"

searching the codebase for now it's used in these cases:

```
{ type: "finite" }
{ type: "nonNaN" }
{ type: "positive" }
{ type: "negative" }
{ type: "nonNegative" }
{ type: "nonPositive" }
{ type: "instanceOf", instanceOf: constructor }
{ type: "trimmed" }
{ type: "pattern", regex }
{ type: "startsWith", startsWith }
{ type: "endsWith", endsWith }
{ type: "includes", includes: searchString }
```
and the new
```
{ id: MinItemsId, minItems }
{ id: MaxItemsId, maxItems }
{ id: ItemsCountId }, itemsCount }
```


so we need to keep in mind that a "filter" annotation can have an `id` but also some associated parameters

jessekelly881: What do you think about doing something like this:

```
/**
 * @category combinators
 * @since 1.0.0
 */
export type AnnotationOptions<A> = {
  tag?: A.Tag
  message?: A.Message<A>
  identifier?: A.Identifier
  title?: A.Title
  description?: A.Description
  examples?: A.Examples
  documentation?: A.Documentation
  jsonSchema?: A.JSONSchema
  [key: string]: any
}

export type FilterAnnotationOptions<A, P extends ReadonlyArray<any>> = AnnotationOptions<A> & {
  props: P
}
```

Adding a tag for `MaxItemsId` sense we might want tags for parsers, transforms, etc., allowing the AnnotationOptions object to be extended with arbitrary props instead of using custom, and having well defined extensions of AnnotationOptions to make these extra props type safe for certain things like filters, parsers, etc.

gcanti: I'm not sure I understood, `tag?: A.Tag` corresponds to what I was calling `id`, but where would the eventual parameters (like the number `maxItems` for the case `MaxItemsId`) go?

jessekelly881: Where MaxItemsId would go and a number like maxItems would go in the props object. Something like:

```js
I.filter((a): a is ReadonlyArray<A> => a.length <= n, { // FilterAnnotationOptions instead of AnnotationOptions type
       tag: MaxItemsId
       description: `an array of at most ${n} items`,
       props: { maxItems: n }
})
```

gcanti: Ok, this is from the perspective of the API for the consumers, but what happens to this information behind the scenes?

Two scenarios come to mind for this:

1) They end up together with `MaxItemsId` in a single "filter" annotation for all filters and `MaxItemsId` serves as a key to identify the specific filter, i.e.:

```
{ 
  [H.ParserHookId]: H.hook(parser),
  [H.ArbitraryHookId]: H.hook(arbitrary),
  ...
  [FilterId]: { [MaxItemsId]: props } 
}
```


2) Or each filter has its own specific annotation whose key corresponds to the ids we are talking about

```
{ 
  [H.ParserHookId]: H.hook(parser),
  [H.ArbitraryHookId]: H.hook(arbitrary),
  ...
  [MaxItemsId]: props
}
```

jessekelly881: What about:

```
{ 
  [H.Tag]: MaxItemsId,
  [MaxItemsId]: { props },

  [H.ParserHookId]: H.hook(parser),
  [H.ArbitraryHookId]: H.hook(arbitrary),
  ...
}
```

Something similar to like Either which is 
`{ _tag: "Right", right: _ }, { _tag: "Left", left: _ }`

gcanti: Ok so in general would be:

```ts
// src/annotation/AST.ts
export const TypeId = "@fp-ts/schema/annotation/TypeId"
```

```ts
// myfilter.ts
export const MyTypeId = "unique/identifier/to/my/Type"

{
  [TypeId]: MyTypeId, // <= official annotation exported by @fp-ts/schema
  [MyTypeId]: { ...MyType params... } // <= namespace for my custom type
}
```

jessekelly881: That would be my guess for the best way for handling it, yeah. What do you think?

jessekelly881: We could probably export some kind of type map that maps the value of TypeId to the type of [MyTypeId]: { ...MyType params... }

jessekelly881: Also, if we had a fully typed ast, the the custom/namespaced arguments could be typed. Which would make it easier to use in compilers for example.

gcanti: > What do you think?

Yes, I think we can discard the `CustomId` annotation in favor of the `TypeId` annotation which is more precise in its intent.

jessekelly881: Ok! Sounds good!

jessekelly881:
@gcanti I will create a separate PR to update the AnnotationOptions object like discussed. Are you ok with this PR as is with the limited scope to data/ReadonlyArray?

gcanti (approved):

PR #79, opened 2023-02-08
extended data/Number

jessekelly881:
Extended data/Number to include positive, negative, nonPositive, nonNegative. Similar to [zod](https://github.com/colinhacks/zod#numbers).

### DIFF
--- a/.changeset/sweet-peas-remember.md
+++ b/.changeset/sweet-peas-remember.md
@@ -1,0 +1,5 @@
+---
+"@fp-ts/schema": patch
+---
+
+extended data/Number to include positive, negative, nonPositive, nonNegative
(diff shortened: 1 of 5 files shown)


jessekelly881:
I thought about that but my worry was with the error messages. It makes more intuitive sense to see "expected a positive number" vs "expected a number greater than 0". But, we could use them with an overloaded error message.

gcanti:
If we use, for example `greaterThan`, AFAIK we can overload its default description:

```ts
export const greaterThan = <A extends number>(
  min: number,
  annotationOptions?: AnnotationOptions<A>
) =>
  (self: Schema<A>): Schema<A> =>
    pipe(
      self,
      I.filter((a): a is A => a > min, {
        description: `a number greater than ${min}`, // default description
        jsonSchema: { exclusiveMinimum: min },
        ...annotationOptions // <= If this contains a `description`, it overrides the default one
      })
    )
```

jessekelly881:
Yeah, that's a better way of defining it. Updated the PR.

gcanti:
Nice thank you, could you please run `pnpm changeset` and then squash the commits?

changeset will ask you for the level of the change, select `patch` and enter a description like the one in the commit (it is needed for the automatic publishing system)

By the way, this procedure with changeset applies to all repos, just for future reference

PR #80, opened 2023-02-08
chore: update to latest ecosystem packages

tim-smart:


tim-smart:
I had a look at what the dual API migration would take, and I'm not sure if is a good idea with all the `options?: AnnotationOptions`. This would require variants for each constructor.

gcanti:
I also think it's not a good idea, I would leave the APIs as they are.

test/compiler/TypeScript.ts
```diff
@@ -168,9 +174,11 @@ const typeScriptFor = <A>(schema: S.Schema<A>): TypeScript<A> => {
         const id = pipe(
           getIdentifier(ast),
           O.map((id) => ts.factory.createIdentifier(id)),
-          O.getOrThrow(() =>
-            new Error(`cannot find an indentifier for this unique symbol ${String(ast.symbol)}`)
-          )
+          O.getOrElse(() => {
```
gcanti, line 177: I released `@fp-ts/core@0.2.1` which contains `O.getOrThrowWith`.

tim-smart: Nice :)

mikearnaldi:
I'll jump ahead and merge this for now as I need it, we can decide later on the api choices such as dual

PR #84, opened 2023-02-10
switch parser to use effect

mikearnaldi:
This is a first pass that does a 1-1 conversion, further work is needed to optimize the various parsers and restructure options to be environmental

### DIFF
--- a/.changeset/selfish-poems-watch.md
+++ b/.changeset/selfish-poems-watch.md
@@ -1,0 +1,5 @@
+---
+"@fp-ts/schema": minor
+---
+
+Switch parser to use Effect
--- a/.vscode/snippets.code-snippets
+++ b/.vscode/snippets.code-snippets
@@ -1,0 +1,12 @@
+{
+  "Gen Function $": {
+    "prefix": "gg",
+    "body": ["function* ($) {}"],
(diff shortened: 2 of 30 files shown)


jessekelly881:
Hey @mikearnaldi! Sorry for my ignorance, but is the benefit of using Effect here vs Either?

mattiamanzati:
> Sorry for my ignorance, but what's the benefit of using Effect vs Either here? 

Effect shines in interpreters because:
- provides stack safety: deep objects or heavy interpreters may result in stack overflow, Effect solves this neatly
- provides a context type: interpreters may need to accumulate an internal state or pass down data while interpreting the schema, effect's context can provide this
- provides both sync and async interpretation: thanks to its fiber runtime, execution will jump to async next tick only if actually needed
- provides an additional untyped internal exception channel: which is a great tool to jump into failure if an unhandled schema case is encountered, without polluting the typed error channel with an "UnexpectedParserCaseError"

jessekelly881:
Oh this is very interesting. Thanks for the explanation. I've used fp-ts for a long time now but effect-ts is pretty new for me.

mikearnaldi:
superseded by #154

PR #85, opened 2023-02-11
add TypeId and PropsId to AnnotationOptions

jessekelly881:
Adds TypeId and PropsId field to AnnotationOptions and removes CustomId


jessekelly881:
@gcanti What do you think of this?

gcanti:
I think that informations related to a custom type, including any configuration parameters, should end up into its specific annotation, indexed by its `TypeId`

```ts
// src/annotation/AST.ts
export const TypeId = "@fp-ts/schema/annotation/TypeId"
```

```ts
export const PatternId = "@fp-ts/schema/data/String/PatternId"

// annotations
{
  [TypeId]: PatternId // <= its TypeId
  [PatternId]: { regex } // <= specific annotation for the custom type, indexed by its TypeId
  ... other annotations ...
}
```

Regarding the `filter` API, it could be something like:

```ts
{
  type: { id: PatternId, params: { regex } },
  description: `a string matching the pattern ${pattern}`,
  jsonSchema: { pattern },
  ...annotationOptions
}
```

the important thing is that you can easily specify both the id and its parameters, and that this information ends up in the way described above.

gcanti:
> I think that informations related to a custom type, including any configuration parameters, should end up into its specific annotation, indexed by its TypeId

Working on this as I need to release a minor version to update the effect/data dependency

PR #86, opened 2023-02-13
Improve compilers dx, customizability, and safety

jessekelly881:
Refactors the compiler code for Arbitrary (at the moment, more to come) to simplify building compilers(by using an easy to create map from ast to model type), make overloading compilers possible(compilers use a map which can be overloaded for by key to override the behavior of the compiler for an arbitrary ast item), and make compilers safer (returns Either instead of throwing)

Replaces compilers with a simple ADT -> Either<E, A> map which is then used to build the compiler. E.g. 

```js
const compilerMap = (
  fc: typeof FastCheck,
  go: AST.Compiler<string, FastCheck.Arbitrary<any>>
): AST.CompilerASTMap<string, FastCheck.Arbitrary<any>> => ({
  NeverKeyword: () => E.left("cannot build an Arbitrary for `never`"),
  StringKeyword: () => E.right(fc.string()),
  NumberKeyword: () => E.right(fc.float()),
  BooleanKeyword: () => E.right(fc.boolean()),
  Refinement: (ast) => pipe(go(ast.from), E.map((arb) => arb.filter((a) => ast.refinement(a)))),
  Union: (ast) => pipe(ast.types.map(go), all, E.map((arbs) => fc.oneof(...arbs))),
  // etc...
  })
```


gcanti:
> make overloading compilers possible

What's the use case for this? Which part of the current compiler you would change?

jessekelly881:
For Arbitrary it's use is less obvious. But in general I think it's good to support overriding so that you don't have to rewrite the entire compiler to make a small change. You might want to overload how the pretty print compilers renders a boolean for example. Instead of rendering true as "true" you might want it to render it as "True".

jessekelly881:
For Arbitrary it's use is less obvious. But in general I think it's good to support overriding so that you don't have to rewrite the entire compiler to make a small change. You might want to overload how the pretty print compilers renders a boolean for example. Instead of rendering true as "true" you might want it to render it as "True". It's also something that we can support with no real cost to DX.

gcanti:
> You might want to overload how the pretty print compiler renders a boolean for example. Instead of rendering true as "true" you might want it to render it as "True".

Why? Given the purpose of Pretty, this also seems contrived to me.

But in any case, why rewrite the entire compiler?

```ts
const myPrettyFor = <A>(schema: Schema<A>): Pretty<A> => {
  switch (schema.ast._tag) {
    case "BooleanKeyword":
      return myOverride
    default:
      return prettyFor(schema)
  }
}
```

jessekelly881:
I don't think this quite works actually. If a boolean is inside of a tuple, for example, the boolean will be rendered normally within the tuple and not be overridden. The override needs to be called by the compiler for tuple, object, etc.

gcanti:
Ah I see, you want the override to work everywhere, but then in the same way, if you allow configuring the compiler, I'm not sure you can have a consistent output since you can no longer control what the `pretty` API returns.
For example, here:
https://github.com/fp-ts/schema/blob/e3786911b5231505ce3d16e783ed75a5ac342044/src/data/ReadonlySet.ts#L18
there is a call to `P.decode`, whose configuration would be decided independently of your configuration and whose output might not be consistent.

jessekelly881:
Exactly. Hmm. I'm not sure I follow. You can ensure that the pretty compiler always returns the same type. Here is the code for compilers I added. 

```js
export type Compiler<Err, Model> = (ast: AST) => E.Either<Err, Model>

export type CompilerASTMap<Err, Model> = {
  [k in keyof TypeMap]: (ast: TypeMap[k]) => E.Either<Err, Model>
}

export type CompilerMapConstructor<Err, Model, Ctx> = (
  c: Ctx,
  g: Compiler<Err, Model>
) => CompilerASTMap<Err, Model>

```

src/Arbitrary.ts
```diff
@@ -46,194 +80,165 @@ const record = <K extends PropertyKey, V>(
...
+
+    return output
   }
 
-  return go(schema.ast)
+interface Context {
+  fc: typeof FastCheck
 }
+
+const compilerMap = (
```
jessekelly881, line 186: This takes a specific ast(e.g. Tuple), the compiler(recursive, used to return the Model for inner types) and returns a Model. In the case of Arbitrary the Model is FastCheck.Arbitrary<any>.

gcanti:
I took some time to better understand.

Let's say instead of a `switch` I want to do pattern matching on `AST` nodes.

**Harness**

```ts
export type Compiler<A> = (ast: AST.AST) => A

export type Match<A> = {
  [K in AST.AST["_tag"]]: (ast: Extract<AST.AST, { _tag: K }>, compile: Compiler<A>) => A
}

export const getCompiler = <A>(match: Match<A>): Compiler<A> => {
  const compile = (ast: AST.AST): A => match[ast._tag](ast as any, compile)
  return compile
}
```

I can rewrite the Pretty compiler like this

```ts
const match: Match<Pretty<any>> = {
  "TypeAlias": (ast, go) =>
    pipe(
      getHook(ast),
      O.match(
        () => go(ast.type),
        ({ handler }) => handler(...ast.typeParameters.map(go))
      )
    ),
  "Literal": (ast) =>
    make(
      I.makeSchema(ast),
      (literal: AST.LiteralValue): string =>
        typeof literal === "bigint" ?
          `${literal.toString()}n` :
          JSON.stringify(literal)
    ),
  "SymbolKeyword": (ast) => make(I.makeSchema(ast), (s) => String(s)),
  "BooleanKeyword": (ast) => make(I.makeSchema(ast), (s) => String(s)),
  "UniqueSymbol": (ast) => make(I.makeSchema(ast), (s) => String(s)),
  "TemplateLiteral": (ast) => make(I.makeSchema(ast), (s) => String(s)),
  "UndefinedKeyword": (ast) => make(I.makeSchema(ast), () => "undefined"),
  "VoidKeyword": (ast) => make(I.makeSchema(ast), () => "void(0)"),
  "NeverKeyword": (ast) =>
    make(I.makeSchema(ast), () => {
      throw new Error("cannot pretty print a `never` value")
    }) as any,
  "UnknownKeyword": (ast) => make(I.makeSchema(ast), (a) => JSON.stringify(a, null, 2)),
  "AnyKeyword": (ast) => make(I.makeSchema(ast), (a) => JSON.stringify(a, null, 2)),
  "ObjectKeyword": (ast) => make(I.makeSchema(ast), (a) => JSON.stringify(a, null, 2)),
  "StringKeyword": (ast) => make(I.makeSchema(ast), (s) => JSON.stringify(s)),
  "NumberKeyword": (ast) =>
    make(
      I.makeSchema(ast),
      (n) => Number.isNaN(n) ? "NaN" : String(n)
    ),
  "BigIntKeyword": (ast) => make(I.makeSchema(ast), (bi) => `${bi.toString()}n`),
  "Tuple": (ast, go) => {
    const elements = ast.elements.map((e) => go(e.type))
    const rest = pipe(ast.rest, O.map(RA.mapNonEmpty(go)))
    return make(
      I.makeSchema(ast),
      (input: ReadonlyArray<unknown>) => {
        const output: Array<string> = []
        let i = 0
        // ---------------------------------------------
        // handle elements
        // ---------------------------------------------
        for (; i < elements.length; i++) {
          if (input.length < i + 1) {
            if (ast.elements[i].isOptional) {
              continue
            }
          } else {
            output.push(elements[i].pretty(input[i]))
          }
        }
        // ---------------------------------------------
        // handle rest element
        // ---------------------------------------------
        if (O.isSome(rest)) {
          const head = RA.headNonEmpty(rest.value)
          const tail = RA.tailNonEmpty(rest.value)
          for (; i < input.length - tail.length; i++) {
            output.push(head.pretty(input[i]))
          }
          // ---------------------------------------------
          // handle post rest elements
          // ---------------------------------------------
          for (let j = 0; j < tail.length; j++) {
            i += j
            output.push(tail[j].pretty(input[i]))
          }
        }

        return "[" + output.join(", ") + "]"
      }
    )
  },
  "TypeLiteral": (ast, go) => {
    const propertySignaturesTypes = ast.propertySignatures.map((f) => go(f.type))
    const indexSignatureTypes = ast.indexSignatures.map((is) => go(is.type))
    return make(
      I.makeSchema(ast),
      (input: { readonly [x: PropertyKey]: unknown }) => {
        const output: Array<string> = []
        const expectedKeys: any = {}
        // ---------------------------------------------
        // handle property signatures
        // ---------------------------------------------
        for (let i = 0; i < propertySignaturesTypes.length; i++) {
          const ps = ast.propertySignatures[i]
          const name = ps.name
          if (ps.isOptional && !Object.prototype.hasOwnProperty.call(input, name)) {
            continue
          }
          output.push(`${prettyName(name)}: ${propertySignaturesTypes[i].pretty(input[name])}`)
          expectedKeys[name] = null
        }
        // ---------------------------------------------
        // handle index signatures
        // ---------------------------------------------
        if (indexSignatureTypes.length > 0) {
          for (let i = 0; i < indexSignatureTypes.length; i++) {
            const type = indexSignatureTypes[i]
            const keys = I.getKeysForIndexSignature(input, ast.indexSignatures[i].parameter)
            for (const key of keys) {
              if (Object.prototype.hasOwnProperty.call(expectedKeys, key)) {
                continue
              }
              output.push(`${prettyName(key)}: ${type.pretty(input[key])}`)
            }
          }
        }

        return isNonEmpty(output) ? "{ " + output.join(", ") + " }" : "{}"
      }
    )
  },
  "Union": (ast, go) => {
    const types = ast.types.map((m) => [P.is(I.makeSchema(m)), go(m)] as const)
    return make(I.makeSchema(ast), (a) => {
      const index = types.findIndex(([is]) => is(a))
      return types[index][1].pretty(a)
    })
  },
  "Lazy": (ast, go) => {
    const f = () => go(ast.f())
    const get = I.memoize<void, Pretty<any>>(f)
    const schema = I.lazy(f)
    return make(schema, (a) => get().pretty(a))
  },
  "Enums": (ast) => make(I.makeSchema(ast), (sn) => JSON.stringify(sn)),
  "Refinement": (ast, go) => go(ast.from),
  "Transform": (ast, go) => go(ast.to)
}
```

And then redefine the `pretty` API like this

```ts
const compile = getCompiler(match)

/**
 * @category prettify
 * @since 1.0.0
 */
export const pretty = <A>(schema: Schema<A>) => (a: A): string => compile(schema.ast).pretty(a)
```

So far we haven't changed anything except the internal implementation.

Now, to get what you have in mind, would it be enough to export `match`?

```diff
-const match: Match<Pretty<any>> = {
+export const match: Match<Pretty<any>> = {
```

So that in order to define a custom compiler, can I patch the `match` officially exported from `/schema`?

```ts
it("should allow for custom compilers", () => {
  const match: typeof P.match = {
    ...P.match,
    "BooleanKeyword": (ast) => P.make(S.make(ast), (b: boolean) => b ? "True" : "False")
  }
  const go = P.getCompiler(match)
  const pretty = <A>(schema: S.Schema<A>) => (a: A): string => go(schema.ast).pretty(a)
  expect(pretty(S.boolean)(true)).toEqual(`True`)
  const schema = S.tuple(S.string, S.boolean)
  expect(pretty(schema)(["a", true])).toEqual(`["a", True]`)
})
```

Code here https://github.com/fp-ts/schema/tree/86

jessekelly881:
Yep! This is exactly what I had in mind as far as customizing current compilers. I think this works well. Although, I do think that the return type from the compiler should be an either instead of throwing. But what do you think? E.g. 

```js
"NeverKeyword": (ast) => make(I.makeSchema(ast), () => E.left("cannot pretty print a `never` value"))
```

gcanti:
Ok, I cannot predict whether these exports will be used, but I don't see any harm in doing so.

Regarding the return type, in general, it's up to the individual compiler to choose which one it is (the `TypeScript` compiler, for example, doesn't need `Either`, and soon `Parser` will return an `Effect`).

If we're specifically talking about `Pretty`, I don't think we need `Either` since the `never` case should never occur, and we could even choose to return `"<never>"` as a representation instead of throwing.

jessekelly881:
Ok. That makes a lot of sense. I will close this PR then. I think this way of going about it is better.

PR #90, opened 2023-02-16
feat: added data/Bigint and N.between

jessekelly881:
Duplicated the functionality of data/Number to data/Bigint and added between and betweenExclusive filters for both modules. Using data/Bigint must be namespaced due to name collisions with number.

### DIFF
--- a/README.md
+++ b/README.md
@@ -471,6 +471,7 @@ pipe(S.number, S.greaterThan(5));
 pipe(S.number, S.greaterThanOrEqualTo(5));
 pipe(S.number, S.lessThan(5));
 pipe(S.number, S.lessThanOrEqualTo(5));
+pipe(S.number, S.between(-2, 2)); // -2 <= x <= 2
 
 pipe(S.number, S.int()); // value must be an integer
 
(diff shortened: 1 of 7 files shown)


gcanti:
Ok for Bigint.
Regarding "between", what if I want `-2 <= x < 2`? If we cannot find an API that allows for full configuration, I would suggest leaving "between" to be derived by end users by leveraging the existing combinators.

jessekelly881:
Makes sense. That said, the @fp-ts/core/Number package exports a between function that is inclusive. The same argument could be made for this fn. What do you think if we copy that interface and remove the betweenExclusive fns but keep the between fns?

gcanti:
> What do you think if we copy that interface and remove the betweenExclusive fns but keep the between fns? To keep things consistent between libraries

👍

jessekelly881:
Updated

README.md
```diff
@@ -483,12 +484,29 @@ pipe(S.number, S.negative()); // < 0
...
+### Bigint filters
+
+```ts
+import * as B from "@fp-ts/schema/data/Bigint";
+
+pipe(S.bigint, B.greaterThan(5n));
+pipe(S.bigint, B.greaterThanOrEqualTo(5n));
+pipe(S.bigint, B.lessThan(5n));
+pipe(S.bigint, B.lessThanOrEqualTo(5n));
+pipe(S.bigint, B.between(-2n, 2n)); // -2 <= x <= 2
```
gcanti, line 496: nitpick: `-2n <= x <= 2n` (same for the following comments)

jessekelly881: Ah good point

jessekelly881: I updated it

gcanti:
Thank you @jessekelly881

PR #97, opened 2023-02-19
Schema: export getPropertySignatures (∼ zod's shape feature)

gcanti:


jessekelly881:
This is exciting! Would it be possible to create a deep version of this? E.g. shape.innerObj.propA

gcanti:
Maybe it's possible, but I would limit it to just one level without complicating things too much.

PR #99, opened 2023-02-21
fix: enhance union unification for boolean and bigint literals

jfet97:
It correctly unify the union of a boolean literal with the `boolean` primitive, and the union of a bigint with the `bigint` one.


jfet97:
May fix #100

PR #109, opened 2023-02-25
fix errata in jsonSchema annotation for Number.between

### LINKED ISSUE #108: `jsonSchema` annotation for `Number.between` uses `max` for both `min` and `max` fields
This package is shaping up really nicely. The abstractions feel clean and have been simple to use and understand. Here's a small issue I bumped into.

### Current Behavior

Looks like a [recent addition](https://github.com/fp-ts/schema/blob/main/src/data/Number.ts#L133) to the Number module included a typo (`min` field shows `max` instead).

### Expected behavior

Expected the `min` field to reflect the user-provided `min` value

### Reproducible example

N/A

### Suggested solution(s)

I'm opening a PR with a fix, raising this issue first per the instructions in the PR template.

### Additional context

N/A

### Your environment

N/A

| Software      | Version(s) |
| ------------- | ---------- |
| @fp-ts/schema | 0.3.0           |
| TypeScript    |   N/A         |

ahrjarrett:
Fixes #108. This is my first contribution, so please let me know if I missed any steps or if you have any requested changes. 

Thank you for your continued work on the project. If were not for fp-ts, I would have found a new career years ago.

### DIFF
--- a/.changeset/modern-cooks-dress.md
+++ b/.changeset/modern-cooks-dress.md
@@ -1,0 +1,5 @@
+---
+"@fp-ts/schema": patch
+---
+
+fixes jsonSchema annotation bug in data/Bigint and data/Number
--- a/src/data/Bigint.ts
+++ b/src/data/Bigint.ts
@@ -84,7 +84,7 @@ export const between = <A extends bigint>(
       self,
       I.filter((a): a is A => a >= min && a <= max, {
         description: `a bigint between ${min}n and ${max}n`,
-        jsonSchema: { maximum: max, minimum: max },
+        jsonSchema: { maximum: max, minimum: min },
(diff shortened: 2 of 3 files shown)


gcanti:
Thanks @ahrjarrett
- looks like there's the same bug in /data/Bigint.ts, could you please fix that too?
- once finished you can run `pnpm changeset`, select "patch" and add a description (that's for the automatic release process)

ahrjarrett:
@gcanti I think I did that right? Let me know if there's anything else I can do. 

I think the PR template could probably use some updating, but I wasn't sure if it made sense to update that until fp-ts/schema is ready to move over to the effect org

gcanti:
Perfect, thanks

PR #115, opened 2023-02-28
Optimize decoding of unions using a heuristic based on literals

gcanti:


src/Parser.ts
```diff
@@ -408,14 +408,32 @@ const parserFor = <A>(
...
         return make(I.makeSchema(ast), (u, options) => {
           const es: Array<PR.ParseError> = []
 
-          // ---------------------------------------------
-          // compute best output
-          // ---------------------------------------------
-          for (let i = 0; i < types.length; i++) {
-            const t = types[i].parse(u, options)
+          if (typeof u === "object" && u !== null) {
+            for (const key of Reflect.ownKeys(separated.keys)) {
```
TylorS: 🔧 Performance-wise making this a regular for-loop should speed it up a bit more 

See: https://www.measurethat.net/Benchmarks/Show/23809/0/for-loop-vs-for-of-perf-test

IIUC all of the built-in array methods tend to be slower than a user-land implementation because they have additional checks to run through for sparse arrays.

src/Parser.ts
```diff
@@ -408,14 +408,32 @@ const parserFor = <A>(
...
         return make(I.makeSchema(ast), (u, options) => {
           const es: Array<PR.ParseError> = []
 
-          // ---------------------------------------------
-          // compute best output
-          // ---------------------------------------------
-          for (let i = 0; i < types.length; i++) {
-            const t = types[i].parse(u, options)
+          if (typeof u === "object" && u !== null) {
+            for (const key of Reflect.ownKeys(searchTree.keys)) {
```
TylorS, line 416: 🔧 Performance-wise this should likely be switched to a for-loop. At small  sizes <= 100 items it's about 10x faster and trails off at larger sizes due to engine implementation differences. IIUC just about all user-land implementations of array built-ins are fasters due to the built-in having additional checks for sparse arrays.

See: https://www.measurethat.net/Benchmarks/Show/23809/0/for-loop-vs-for-of-perf-test

test/Decoder.ts
```diff
@@ -838,6 +838,14 @@ describe.concurrent("Decoder", () => {
     Util.expectDecodingFailure(schema, 1, "Expected never, actual 1")
   })
 
+  it("union/ members with literals but the input doesn't have any", () => {
+    const schema = S.union(
+      S.struct({ a: S.literal("A"), c: S.string }),
+      S.struct({ b: S.literal("B"), d: S.number })
+    )
+    Util.expectDecodingFailure(schema, {}, "Expected never, actual {}")
```
TylorS, line 846: 🔧 This isn't quite the error message I would expect here. I'd expect a representation of the union instead of `never`

src/Parser.ts
```diff
@@ -482,6 +508,71 @@ const parserFor = <A>(
...
+export const _getSearchTree = <A extends Schema<any>>(
+  schemas: ReadonlyArray<A>,
+  as: "decoder" | "guard" | "encoder"
+): {
+  keys: { [key: PropertyKey]: { [literal: string]: ReadonlyArray<A> } }
+  otherwise: ReadonlyArray<A>
+} => {
+  const keys: { [key: PropertyKey]: { [literal: string]: Array<A> } } = {}
+  const otherwise: Array<A> = []
+  for (const schema of schemas) {
```
TylorS, line 561: 🔧 perf-wise a regular for-loop will do a bit better

src/Parser.ts
```diff
@@ -428,7 +454,7 @@ const parserFor = <A>(
           // ---------------------------------------------
           return I.isNonEmptyReadonlyArray(es) ?
             PR.failures(es) :
-            PR.failure(PR.type(AST.neverKeyword, u))
+            PR.failure(PR.type(AST.neverKeyword, input))
```
TylorS, line 457: ❓ Would the following not work here for a better error message? `PR.failure(PR.type(ast, input))`

gcanti: Yes, that branch was impossible to reach (so never was ok), but now it has become possible. Yesterday I started making the change you propose, but then I stopped because the /formatter/Tree/formatExpected function should probably be improved as well, and before making any other changes, I wanted to be sure that the algorithm was correct.

I made the change in the last commit to show the result.

TylorS: Okay cool, that makes sense. I was just thinking the same thing about the formatter when I saw `type literal or type literal` in the test.

I have some ideas about the formatting, but I suspect you're also a step ahead of me there. Have you given thought to how you'd improve type literal formatting?

gcanti: > Have you given thought to how you'd improve type literal formatting?

Not yet (I was distracted by the fp-ts -> effect transfer), I'm open to suggestions!

TylorS: The current implementation of `getExpected` uses annotations for Identifier, Title, and Description as possible places to print information with. So my first thought was to make AST.TypeLiteral to append these annotations as well in a concatenated form. This definitely has some runtime cost as it would need to iterate each property/index signature to create them and also the checks for the annotations to begin with, which could also couple its implementation to a specific set of "well known" annotations.

The next thing, which I think is better, that came to mind was building the compiler/interpreter much closer to say Pretty. Pretty utilizes a hook for further customization as necessary which keeps the implementation lazy and generalized.

gcanti: Using `ast` is a safe bet but also means that the error message, even if imroved, will be highly generic and the user will be forced to dig out the reason why decoding failed by himself. Alternatively we could leverage the existing checks in order to provide more meaningful errors, I committed a POC, example:
```ts
const schema = S.union(
  S.struct({ a: S.literal(1), c: S.string }),
  S.struct({ b: S.literal(2), d: S.number })
)
Util.expectDecodingFailure(schema, null, "Expected type literal, actual null")
Util.expectDecodingFailure(schema, {}, "/a is missing, /b is missing")
Util.expectDecodingFailure(schema, { a: null }, `/a Expected 1, actual null, /b is missing`)
Util.expectDecodingFailure(schema, { b: 3 }, `/a is missing, /b Expected 2, actual 3`)
```

TylorS: I'm not sure why it'd necessarily be too generic. If `S.string` could be understood to create `string`, and `S.literal(1)`  to create `1` and `S.struct` was concatenated its property signatures I think printing 
`expected { a: 1, c: string }, actual null` would be a lot more helpful.

This could be expanded pretty far I think, refinement would have to know how to append its constraints on the underlying schema, transform could use from ast.from, etc.

As a slight tangent to this, ParseResult could be missing a `Wrap` style error like in io-ts. It's really helpful to be able to add additional context to errors, and maybe that's an easy workaround. Something like the following would immediately make this error more useful IMO.

```
 MyUnionName
└─ Expected type literal, actual null
```

TylorS: After writing that, I realize it's basically the TypeScript compiler in the test folder, but straight to strings instead of constructing an AST in between.

gcanti: The current implementation looks more consistent with all the other error messages to me, for example:

```ts
S.decodeOrThrow(schema)({ a: null })
/*
2 error(s) found
├─ key "a"
│  └─ Expected 1, actual null
└─ key "b"
   └─ is missing
*/
```

the error message explains the actual reasons why the decoding failed.

With `ast` you'd get something like

```ts
S.decodeOrThrow(schema)({ a: null })
/*
1 error(s) found
└─ Expected { a: 1, c: string, ...other props... } or { b: 2, d: number, ...other props... } or ... other members..., actual {"a":null}
*/
```

and then the user must parse the types in its head and realize that:
- for the first member the problem is: the "a" key has values null but should be 1
- for the second member the problem is: the "b" key is missing 
...
- for the n-th member the problem is: ...

TylorS: Oh I 100% agree what's existing is a whole lot nicer in these examples. But it's not very helpful in this case

```ts
S.decodeOrThrow(schema)(null)
/*
2 error(s) found
├─ union member
│  └─ Expected type literal, actual null
└─ union member
   └─ Expected type literal, actual null 
*/

Maybe there's another way to improve that specific situation than creating a new hook
```

gcanti: with `null` I get

```
1 error(s) found
└─ Expected type literal, actual null
```

TylorS: Same here. I'm using the published version though - https://codesandbox.io/s/mystifying-goldwasser-tyskzy?file=/src/index.ts

TylorS: But I guess mostly my point is having _something_ identifying the schema instead of just `type literal` in those error messages

src/Parser.ts
```diff
@@ -586,15 +588,25 @@ export const _getSearchTree = <A extends Schema<any>>(
...
+        const [key, literal] = tags[j]
+        const hash = String(literal)
+        keys[key] = keys[key] || { buckets: {}, literals: [] }
+        const buckets = keys[key].buckets
+        if (Object.prototype.hasOwnProperty.call(buckets, hash)) {
+          if (j < tags.length - 1) {
+            continue
+          }
+          buckets[hash].push(member)
+          keys[key].literals.push(AST.createLiteral(literal))
```
TylorS, line 603: 🔧 It's a pretty minor performance thing, but if `_getLiterals` returned the full literal AST node and `hash = String(literal.literal)` we'd avoid recreating the literal nodes here and in the else clause

src/Parser.ts
gcanti: > I'm using the published version though

Ah ok, in the example above I was referring to this PR.

> But I guess mostly my point is having something identifying the schema instead of just type literal in those error messages

Yeah I agree 👍 , the error messages can definitely be improved (keep in mind that I didn't work on them much because I needed them primarily for testing). What matters most to me in this PR concerning unions is getting the algo and the underlying error model right, as for how they will be formatted, I think that can be addressed in a separate issue / PR.

TylorS: Agreed. I left one more minor perf suggestion -- you can take it or leave it

TylorS (approved):
Looks great! The performance of decoding unions with literals has definitely increased quite a bit

PR #116, opened 2023-02-28
support getters on prototype chain when decoding

tim-smart:
Should decoding getters work? I think it should.

### DIFF
--- a/src/Parser.ts
+++ b/src/Parser.ts
@@ -307,7 +307,7 @@ const parserFor = <A>(
               const parser = propertySignaturesTypes[i]
               const name = ps.name
               expectedKeys[name] = null
-              if (!Object.prototype.hasOwnProperty.call(input, name)) {
+              if (!I.hasOwn(input, name)) {
                 if (!ps.isOptional) {
                   const e = PR.key(name, [PR.missing])
                   if (allErrors) {
@@ -420,7 +420,7 @@ const parserFor = <A>(
                 const name = ownKeys[i]
                 const buckets = searchTree.keys[name].buckets
                 // for each property that should contain a literal, check if the input contains that property
-                if (Object.prototype.hasOwnProperty.call(input, name)) {
+                if (I.hasOwn(input, name)) {
                   const literal = String(input[name])
                   // check that the value obtained from the input for the property corresponds to an existing bucket
                   if (Object.prototype.hasOwnProperty.call(buckets, literal)) {
--- a/.changeset/brave-timers-call.md
+++ b/.changeset/brave-timers-call.md
@@ -1,0 +1,5 @@
+---
+"@fp-ts/schema": patch
+---
+
+include getters from prototype chain when decoding
--- a/src/internal/common.ts
+++ b/src/internal/common.ts
@@ -306,6 +306,15 @@ export const getTemplateLiteralRegex = (ast: AST.TemplateLiteral): RegExp => {
 // ---------------------------------------------
 
 /** @internal */
+export const hasOwn = (o: object, prop: PropertyKey): boolean => {
+  if ("__proto__" in o && o.__proto__ !== Object.prototype) {
+    return Object.prototype.hasOwnProperty.call(o, prop) || hasOwn(o.__proto__ as object, prop)
+  }
+
+  return Object.prototype.hasOwnProperty.call(o, prop)
+}
+
+/** @internal */
 export const memoize = <A, B>(f: (a: A) => B): (a: A) => B => {
   const cache = new Map()
   return (a) => {
--- a/test/Decoder.ts
+++ b/test/Decoder.ts
@@ -833,6 +833,66 @@ describe.concurrent("Decoder", () => {
     )
   })
 
+  it("struct/ constructor", () => {
+    const schema = S.struct({ constructor: S.any })
+    Util.expectDecodingSuccess(
+      schema,
+      { constructor: "a" }
+    )
+    Util.expectDecodingFailure(
+      schema,
+      {},
+      "/constructor is missing"
+    )
+  })
+
+  it("struct/getters", () => {
+    const schema = S.struct({ name: S.string })
+    class WithName {
+      get name() {
+        return "Tim"
+      }
+    }
+    Util.expectDecodingSuccess(
+      schema,
+      new WithName(),
+      { name: "Tim" }
+    )
+  })
+
+  it("struct/getters unexpected", () => {
+    const schema = S.struct({ name: S.string })
+    class WithUnexpected {
+      get name() {
+        return "Tim"
+      }
+
+      unexpected = "fail"
+    }
+    Util.expectDecodingFailure(
+      schema,
+      new WithUnexpected(),
+      `/unexpected is unexpected`
+    )
+  })
+
+  it("struct/getters additional", () => {
+    const schema = S.struct({ name: S.string })
+    class WithUnexpected {
+      get name() {
+        return "Tim"
+      }
+      get ignore() {
+        return "ignore"
+      }
+    }
+    Util.expectDecodingSuccess(
+      schema,
+      new WithUnexpected(),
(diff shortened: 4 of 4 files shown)


tim-smart:
Seems this also fixes an issue in /match, when matching against things like `Response` from the fetch API - which is a class with getters.

src/Parser.ts
```diff
@@ -309,7 +309,7 @@ const parserFor = <A>(
               const parser = propertySignaturesTypes[i]
               const name = ps.name
               expectedKeys[name] = null
-              if (!Object.prototype.hasOwnProperty.call(input, name)) {
```
gcanti, line 312: If we consider other keys than own keys, I think we can use `name in input` here (and consequently no need for a new `hasOwnProp` function).

Note that there are other 2 calls like `Object.prototype.hasOwnProperty.call(input, name)` (another one in `Parser` and one in `Pretty`), should be changed as well I guess

gcanti (changes requested):

gcanti:
> Should decoding getters work? I think it should.

In general we should come up with a possibly objective rule in order to answer this kind of questions.

IMO the answer should derive from the TypeScript behaviour itself.

Since the following is allowed by TypeScript:

```ts
class WithName {
  get name() {
    return "Tim"
  }
}

type A = { name: string }

export const a1: A = new WithName() // ok
```

I would agree that the answer should be positive.

src/Parser.ts
tim-smart: If we use `name in input` then you will also get all these as candidates:

```
> Object.getOwnPropertyNames({}.__proto__)
[
  'constructor',
  '__defineGetter__',
  '__defineSetter__',
  'hasOwnProperty',
  '__lookupGetter__',
  '__lookupSetter__',
  'isPrototypeOf',
  'propertyIsEnumerable',
  'toString',
  'valueOf',
  '__proto__',
  'toLocaleString'
]
```

I think doing this could introduce subtle bugs in programs?

gcanti: Could you provide an example?

tim-smart: Probably the most probable one I can think of:

```ts
const schema = Schema.struct({ constructor: S.any })

// returns true
Parser.is(schema)({})
```

gcanti: Seems a bit contrived. Also TypeScript is ok with it

```ts
type A = { constructor: any }
const a: A = {} // ok
```

tim-smart: Yes it is a bit contrived, but not impossible!

Another thing just came to mind, which I have addressed in the latest commit, is how to handle `isUnexpectedAllowed`.

We definitely don't to return errors for the above property names :)

gcanti: It's a bit contrived for the use of the `S.any` type, not so much for the `constructor` property. Not only is it possible, but it's also legitimate, as suggested by the example with TypeScript. So I haven't yet understood what the problem is with `name in input`. If a user defines a schema as `S.struct({ constructor: S.any })`, it means they're interested in obtaining the `constructor` property, and therefore we provide it accordingly. Where would the problem be?

tim-smart: Let's say I have a car API that has a `constructor` field as input, and for some reason or another I type it as `unknown` or `any`:

```ts
const input = S.struct({ model: S.string, constructor: S.unknown })
```

If I forget to add the "constructor" field, it will still be flagged a valid input:

```json
{ "model": "Ford" }
```

When I should have sent something like:
```json
{ "model": "Ford", "constructor": { "name": "asdf" } }
```

gcanti: Another way to establish what and where to search, it is to mimic what destructuring does:

```ts
const _a = Symbol.for("a")
const _b = Symbol.for("b")
class A {
  a = 1;
  [_a] = 2
}
class B extends A {
  b = 3;
  [_b] = 4

  get c() {
    return 5
  }

  d(): number {
    return 6
  }
}

const { a, c, constructor, ...rest } = new B()
console.log(a, c, constructor, rest) // 1 5 [class B extends A2] { b: 3, [Symbol(a)]: 2, [Symbol(b)]: 4 }
```

where

- `a`, `c`, `constructor` represent the keys of the schema defined by the user (e.g. `S.struct({ a: S.any, c: S.any, constructor: S.any })`)
- `rest` represents where to search for unexpected keys

which leads to the following:

- we search `a`, `c`, `constructor` using `name in input` (the only change to make)
- we search for unexpected keys only in own properties (as it is now)

gcanti: Here's some tests https://github.com/Effect-TS/schema/compare/main...fix/116

tim-smart: I think if I were to make a runtime rule for this, it would be:

> Reflect.ownKeys up to and excluding Object.prototype.

Implementation:

```ts
export function ownKeys(o: object): ReadonlyArray<string | symbol> {
  if ("__proto__" in o && o.__proto__ !== Object.prototype) {
    return Reflect.ownKeys(o).concat(ownKeys(o.__proto__ as object))
  }

  return Reflect.ownKeys(o)
}
```

This will mean:

- `{}` correctly has no valid keys
- classes and classes that extend other classes will have all their properties considered
- non-objects / primitives will also have no valid keys (`const {constructor} = "hello"` is valid 😅, `"constructor" in "hello" correctly throws though!)

For individual property checks, something like:

```ts
export function hasOwn(o: object, prop: string | symbol): boolean {
  if ("__proto__" in o && o.__proto__ !== Object.prototype) {
    return Object.hasOwn(o, prop) || hasOwn(o.__proto__ as object, prop)
  }

  return Object.hasOwn(o, prop)
}
```

gcanti: > Reflect.ownKeys up to and excluding Object.prototype.

Okay, but why? There are many possible rules that we can come up with. To avoid complete arbitrariness, I think it's better to refer to some language construct (in JavaScript or TypeScript), such as in the case of destructuring, so that the library's behavior is more intuitive for users and easier to explain.

tim-smart: I think the exclusion of `Object.prototype` is important, because it represents the boundary between properties defined with intention vs properties outside of your control (in most cases).

It also means {} has no valid properties, which I think most users of the library would expect.

If we use the `in` operator, then {} has valid properties as mentioned above.

gcanti: The boundary between properties that are defined with intention versus properties outside of your control is established by the user when they express that property in their schema, just as they do during destructuring:

```ts
//      v-- expected keys ----v (explicitly expressed by the user, search in own properties + prototype chain)
const { a,    c,    constructor, ...rest } = input
//                                   ^-- unexpected keys (search in own properties)
```

^ Here the user explicitly wants to obtain the `constructor` property for their own reasons that we cannot exclude a priori, so why should we interfere in /schema when faced with the same situation?

gcanti: > a user might actually want to accept something from Object.prototype

Yeah, and from my point of view we are not "including", but rather "not excluding".
What is "valid" or not should be determined as much as possible by the user, not by us.

tim-smart: What would think about using `ownKeys` by default, then adding `includePrototypeChain` to the Parser options?

gcanti: My preference goes to mimicking the destructuring behaviour without configuration required

gcanti: > My preference goes to mimicking the destructuring behaviour with no configuration required

However, given that any decision is arbitrary to some extent, I'm okay with asking for other options. No problem for me.

datner:
I agree with @tim-smart for the reason @gcanti explained above

datner:
> Adding support for getters is fine, there is disagreement about the way to do it though https://github.com/Effect-TS/schema/pull/116#discussion_r1126103412

I saw, but I don't have a strong opinion on the methods. I am not versed enough in the trade-offs so I don't want to muddy the water with an empty opinion

PR #129, opened 2023-03-03
Add shell hook to default Nix devShell

IMax153:


mikearnaldi:
I've recently removed the `pnpm i & pnpm run build` steps because they are noisy with direnv and we would use vscode tasks instead

PR #150, opened 2023-03-15
Add Schema namespace

tim-smart:
Just a suggestion, to be more consistent with the effect ecosystem. I.e. `Effect.Success`, `Config.Primitive` etc.


mikearnaldi:
While it would be more standard I think the special case in schema is that inferring the type from value is far more common so I'd avoid using namespacing here

PR #161, opened 2023-03-20
Chore/query

mikearnaldi:


docs/modules/AST.ts.md
```diff
@@ -967,7 +978,7 @@ Added in v1.0.0
 **Signature**
 
 ```ts
-export interface Transform extends Annotated, HasTransformation {
```
datner, line 970: good, `HasTransformation` is not an ancestor, it's a trait 💯

PR #163, opened 2023-03-21
feat: Renamed `isUnexpectedAllowed` to `allowExcess`

OlaoluwaM:
- This change is motivated by `allowExcess` aligning more, in name, with the use case for the `isUnexpectedAllowed` option
- This change also extends to the `Unexpected` parser error, changing it to `Excess` for consistency


OlaoluwaM:
I think you mean the flag shouldn't be true by default right?

datner:
`allowedExcess` is not really common language in this sphere. I don't care about the naming, I think that `isX` is the best boolean-indicating language despite it not being super natural. I would use `isUnexpectedAllowed` for a default-false-optional-true property. I would use `allowUnexpected` for a compound property such as `boolean | 'setting-a' | 'setting-b'`.
Spilling over to a different proposal, I would also name it according to its default. A default-true-optional-false should be called `isStrictShape` or something like that

jessekelly881:
Currently when using isUnexpectedAllowed: true the extra properties are stripped by from the type and at runtime so that's not an issue. As for prefixing with is, in general I agree, booleans and predicates should probably be prefixed with is. The only reason I recommended "strict" without the is is because it matches the way tsconfig names it.

datner:
I also do think that `unexpected` is a more accurate term than `excess`. Excess represents more-than-asked, or above-expectations, so it is accurate. But something that is Unexpected represents astonishment or surprise. The former being more _technically_ accurate, which is good. But the latter is more _semantically_ accurate. Humans are semantic creatures, so the technical excess is not note-worthy, whereas getting more than you bargained for is a potentially note-worthy event

Compare overfilling a bucket with **excess** water to finding an **unexpected** frog in the bucket.

As noted by @patroza, but unlike @jessekelly881 opinion. I agree that there should be no spill-over of properties, but I do think that strict-by-default is uncomfortable now and life-saving later. Intentionally opting-out of safety should be the choice, not intentionally posing restrictions. Note the mental positioning change, since I can't imagine a situation where I turn this flag on by choice.

datner:
>The only reason I recommended "strict" without the is is because it matches the way tsconfig names it.

strict is better than `allowX` and similar verbs since it's established. Though I'd watch out for conflating terminology. Thinking back to the confusion `zod` caused by adopting real-but-opaque terminology from typescript (`extends`, `intersection`) and changing the mental model associated with them, _projecting back_ on the original subject from which the terminology was originally adopted ;)

patroza:
@datner as always enjoy reading you. I would suggest that allowing additional properties has little to do with safety, but more with convenience. The decoder is the gatekeeper and leaving all the unwanted stuff outside. I think that's the right default. It is only mildly useful, in very specific cases, to error on receiving more than you asked for, imo.

jessekelly881:
I agree @patroza. I can't think of many cases where you would want to error on receiving extra/unexpected properties when the parser narrows that down into what you were expecting either way. What comes out of S.parse is the same whether there are excess properties or not.

datner:
> yes, but I iterate over them after the gate, not before :) 

Yeah ofc, I meant that removing the properties _is_ a runtime safety concern and not just convenience 😅 

My vote is erring on the side of strictness. `isUnexpectedAllowed: false` helps you discover unexpected data. Especially useful when working with application edges. For example, I had no idea that in runtime I'm leaking a property when saving some entity to the database. I had type safety, but I had no runtime safety. The parser might have stripped the properties away, but them being there at all is a bug that I should fix and not just cross my fingers hoping that every other person using this property has no plans anywhere near those leaked properties.

Or in shorter terms: I wanna know if I fucked up, don't hide it from me just because it'll be a bit uncomfortable

patroza:
> Or in shorter terms: I wanna know if I fucked up, don't hide it from me just because it'll be a bit uncomfortable
I think a `paranoid` mode is better for this. There are imo many cases where over providing is totally valid, especially when coming from `unknown`.

jessekelly881:
Seconded. If I'm working with data from an api, for example, I want to be able to access certain props in a typesafe way but I don't care what else the api returns and I certainly don't want my application to break every time an additional prop is added to the api.

patroza:
Distinguishing by boolean feels like an internal api anyway to me, I think as a public api, you could expose `parse` and `parseStrict`/Paranoid or so.

jessekelly881:
That seems better, yeah.

datner:
> I think a `paranoid` mode is better for this. There are imo many cases where over providing is totally valid, especially when coming from `unknown`.

Eh, tbh I agree with `paranoid` mode. Just like theres `strict` settings and `strictest` settings so many ts project setup tools, I concede my point

> I certainly don't want my application to break every time an additional prop is added to the api.

Funny you say that. It's exactly what happened to me 2 weeks ago, but that actually was a boon because the change broke me in subtle way that I would have hated to figure out later

>Distinguishing by boolean feels like an internal api anyway to me, I think as a public api, you could expose parse and parseStrict/Paranoid or so.

I was just about to suggest this. Remove the option and just express it as a method.

mikearnaldi:
regarding defaults I do think the default should allow for excess properties, reading the past history of this chat there are reference to "safety" and "deleting properties" that isn't what's happening at all, there are 2 settings one allows for unexpected properties to be present the other makes it error in case there are unexpected properties, the output is always the same regardless and the input object is never mutated.

regarding naming conventions, `isUnexpectedAllowed` looks terrible and generally speaking I don't think it makes any sense to buy conventions without context like "all boolean should have prefix is", sounds to me like "all types should begin with T", neither makes sense.

I would also avoid exposing different methods, in TS land options are usually object arguments and people are familiar with those adding methods would expose us to exponential method growth as any we would need every possible combo

OlaoluwaM:
@gcanti I think if there are only going to be two options, leaving it as a boolean with a more semantic name should have the same effect. Thoughts?

gcanti:
"Excess" like in *excess property checking* https://www.typescriptlang.org/docs/handbook/interfaces.html#excess-property-checks

OlaoluwaM:
Hmm, @datner makes a point. The verbose options _are_ less ambiguous and can be easily extended. Though, for "all or nothing" cases, I think booleans would work just fine, unless there is an intent for extension later on

jessekelly881:
Verbose options seems much better. onUnexpected seems to imply that it's being handled on a case by case basis vs onExcess which seems to imply that the excess props are handled as a group. I would intuitively expect onUnexpected to take a function as an argument that would allow me to handle extra props on a case by case basis, for example.

datner:
@gcanti That page is now deprecated and in the new docs the word `excess` is basically non-existent 🤡 
But I understand that it is the 'official' name for properties-that-shouldn't-be-there, still I think it's a bad name

PR #164, opened 2023-03-21
chore: Upgrade to @effect/data 0.5.0

TylorS:
Simply brings versions inline with the rest of the Effect ecosystem


mikearnaldi:
0.3 out

PR #173, opened 2023-03-22
feat: added S.required combinator

jessekelly881:
Adds S.required

```js
const schema = S.required(S.struct({ a: S.optional(S.number) }))

expectParseSuccess(schema, { a: 1 }, { a: 1 })
expectParseFailure(schema, {}, "/a is missing")
```

https://github.com/colinhacks/zod#required

### DIFF
--- a/README.md
+++ b/README.md
@@ -866,6 +866,13 @@ pipe(S.struct({ a: S.string, b: S.number }), S.omit("a"));
 S.partial(S.struct({ a: S.string, b: S.number }));
 ```
 
+## Required
+
+```ts
+// $ExpectType Schema<Required<{ readonly a?: string; readonly b?: number; }>>
+S.required(S.struct({ a: S.optional(S.string), b: S.optional(S.number) }));
+```
+
 ## Records
 
 ### String keys
--- a/src/AST.ts
+++ b/src/AST.ts
@@ -1028,6 +1028,56 @@ export const partial = (ast: AST): AST => {
   }
 }
 
+/**
+ * Equivalent at runtime to the built-in TypeScript utility type `Required`.
+ *
+ * @since 1.0.0
+ */
+export const required = (ast: AST): AST => {
+  switch (ast._tag) {
+    case "Tuple": {
+      const restElement = pipe(
+        ast.rest,
+        O.map((rest) => createUnion(rest))
+      )
+
+      return createTuple(
+        [
+          ...ast.elements.map((e) => createElement(e.type, false)),
+          ...(O.isSome(restElement) ? [createElement(restElement.value, false)] : [])
+        ],
+        pipe(
+          ast.rest,
+          O.map((rest) => [createUnion(rest)])
+        ),
+        ast.isReadonly
+      )
+    }
+    case "TypeLiteral":
+      return createTypeLiteral(
+        ast.propertySignatures.map((f) =>
+          createPropertySignature(
+            f.name,
+            f.type,
+            false,
+            f.isReadonly,
+            f.annotations
+          )
+        ),
+        ast.indexSignatures
+      )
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -866,6 +866,13 @@ pipe(S.struct({ a: S.string, b: S.number }), S.omit("a"));
 S.partial(S.struct({ a: S.string, b: S.number }));
 ```
 
+## Required
+
+```ts
+// $ExpectType Schema<Required<{ readonly a: string; readonly b: number; }>>
```
gcanti, line 872: typo, should be `Schema<Required<{ readonly a?: string; readonly b?: number; }>>`

src/AST.ts
```diff
@@ -1028,6 +1028,47 @@ export const partial = (ast: AST): AST => {
...
+ * @since 1.0.0
+ */
+export const required = (ast: AST): AST => {
+  switch (ast._tag) {
+    case "Tuple":
+      return createTuple(
+        ast.elements.map((e) => createElement(e.type, false)),
+        pipe(
+          ast.rest,
+          O.map((rest) => [createUnion(rest)])
```
gcanti, line 1043: is this right?

in

```ts
type A = readonly [string, ...Array<number>, boolean]
type B = Required<A>
```

`B` is

```ts
type B = readonly [string, ...(number | boolean)[], number | boolean]
```

not 

```ts
type B = readonly [string, ...(number | boolean)[]]
```

src/AST.ts
```diff
@@ -1028,6 +1028,47 @@ export const partial = (ast: AST): AST => {
...
+            f.annotations
+          )
+        ),
+        ast.indexSignatures
+      )
+    case "Union":
+      return createUnion(ast.types.map((member) => required(member)))
+    case "Lazy":
+      return createLazy(() => required(ast.f()))
+    case "Refinement":
```
gcanti, line 1064: I'm not sure we can actually handle refinements or transformations, in `partial` I'm throwing at the moment

gcanti (changes requested):

README.md
jessekelly881: Fixed

src/AST.ts
jessekelly881: Oh good catch. Let me see if I can fix it.

jessekelly881: The required type on tuples is odd.. For example,

```js
type A = readonly ["1", ...Array<"2">, "3", ...Array<"4">];
type B = Required<A>;

const b: B = ["1"]; // type checks correctly
```

src/AST.ts
jessekelly881: Hmm. These tests are passing for me: 

```js
  it("required/ filter", () => {
    const schema = S.required(S.struct({
      a: S.optional(pipe(S.number, S.greaterThan(0)))
    }))

    expectParseSuccess(schema, { a: 1 }, { a: 1 })
    expectParseFailure(schema, {}, "/a is missing")
  })

  it("partial/ filter", () => {
    const schema = S.partial(S.struct({
      a: S.optional(pipe(S.number, S.greaterThan(0)))
    }))

    expectParseSuccess(schema, { a: 1 }, { a: 1 })
    expectParseSuccess(schema, {}, {})
  })
```

src/AST.ts
jessekelly881: Typescript is perfectly happy to have a rest element after a rest element, as strange as it behaves.. but schema isn't.

jessekelly881: That said, I think I've fixed it.

src/AST.ts
gcanti: ok merging, I'll check out later, thanks

PR #192, opened 2023-03-28
feat: FromOptionalSchema + optionFromOptional

tim-smart:


tim-smart:
@gcanti Let me know if you want to approach this differently or don't think it is a good idea.

I guess another way of doing it, is to explicitly transform from isOptional false -> true first.

Could also add the default value on to FromOptionalSchema to reduce the extra transform for withDefault.

gcanti:
ideally we should handle optionals with a single API (`optional`) 

also it should play well with other combinators, in particular `extend`, but I feel like the way we are modeling transformations as an opaque pair decode/encode makes things quite hard

tim-smart:
OK I'll see what we can do to reduce the API surface.

tim-smart:
Got it working with `extend`, which required modification at the AST level. I ended up using an annotation as it doesn't seem to be something representable at the type level.

I kept the `withDefault` combinator, as it can then have a dual API where as `optional` can only be used schema-first.

gcanti:
Adding something to the AST or adding a new annotation lead to the same result, you are implicitly defining a transformation from an optional prop to a required prop, but this isn't going to work because then `isOptional` doesn't make sense anymore

gcanti:
Not yet, I've been thinking about this for a couple of weeks and I didn't come up with anything actionable (and consistent with other APIs like extend, pick, omit, getPropertySignatures, etc...)

tim-smart:
OK I'll move this to a draft PR and let you think about it some more :)

My only other idea was making `isOptional` an enum to detect "to required", and forcing the schema to handle `undefined`.

gcanti:
@tim-smart I wrote a POC in the `next` branch.

Added a `ReadonlyArray<PropertySignatureTransformation>` field to the `Transform` AST node representing a list of `PropertySignature -> PropertySignature` transformations.

In `test/PropertySignatureTransformations.ts` there are some examples of possible transformations (besides "default" and "Option").

Also changed the `optional` constructor to accept a configuration (examples in `test/optional.ts`).

The `extend` API is compatible, meaning that for example you can do

```ts
/*
const schema: S.Schema<{
    readonly b: string;
    readonly a?: string;
    readonly d: boolean;
    readonly c?: number;
}, {
    readonly a: string;
    readonly b: string;
    readonly c: number;
    readonly d: boolean;
}>
*/
const schema = pipe(
  S.struct({ a: S.optional(S.string, { to: "default", value: "" }), b: S.string }),
  S.extend(S.struct({ c: S.optional(S.number, { to: "default", value: 0 }), d: S.boolean }))
)
```

( examples in `test/extends.ts`)

tim-smart:
Looks pretty good!

My only thought is to make default values lazy. One reason for this is constructing dates on creation.

PR #218, opened 2023-04-05
feat: Added missing assertion functions for some AST nodes

OlaoluwaM:
Going through the code, I noticed some guards were missing, so I added them :)


gcanti:
Thanks @OlaoluwaM, could you please run `pnpm changeset`, select "patch" and add a description (the commit message would be ok), that's for the automatic release process

PR #220, opened 2023-04-05
update dtslint

gcanti:


jessekelly881:
Have you seen that vitest has type checking?

gcanti:
does it run the tests against a range of typescript versions?

jessekelly881:
That.. is a good question. They don't mention it in their docs.

PR #231, opened 2023-04-14
feat: isValidDate filter

tim-smart:


gcanti:
@tim-smart we could fix the issue in [`@effect/data/Predicate`](https://github.com/Effect-TS/data/blob/6c33bb084508f44a6ee8c7900c2cf85ae79a367e/src/Predicate.ts#L366):

```diff
-export const isDate = (input: unknown): input is Date => input instanceof Date
+export const isDate = (input: unknown): input is Date => input instanceof Date && !isNaN(input.getTime())
```

gcanti:
Alternatively, if we don't want to touch effect/data, we can fix the `date` schema definition:

```ts
const isDate = (input: unknown): input is Date => Predicate.isDate(input) && !isNaN(input.getTime())

/**
 * @category constructors
 * @since 1.0.0
 */
export const date: Schema<Date> = declare(
  [],
  struct({}),
  () => (u) => !isDate(u) ? PR.failure(PR.type(date.ast, u)) : PR.success(u),
  {
    [AST.IdentifierAnnotationId]: "Date",
    [I.PrettyHookId]: datePretty,
    [I.ArbitraryHookId]: dateArbitrary
  }
)
```

jessekelly881:
Hmm. It seems like there are two distinct things being tested for here. 1) whether the object is an instance of Date and 2) whether the instance of Date is valid. Maybe having a distinct _isDateValid_ filter is the way to go?

gcanti:
Yeah, adding a filter and changing `dateFromString`'s transformation target from `date` to `pipe(date, theNewfilter())` looks good

tim-smart:
Should be add isValidDate to Predicate?

tim-smart:
@gcanti Predicate doesn't seem like the right place, so implemented it only in /schema.

jessekelly881:
I think it mades sense to create an @effect/data/Date module and add it there.

PR #237, opened 2023-04-17
export ValidDateFromSelf and rename validDate filter

gcanti:
@tim-smart renamed `isValidDate` to `validDate` for consistency with other filters.

In the next minor I think we should also rename `date` to `DateFromSelf` and `DateFromString` to `Date`


tim-smart:
I think dateFromSelf / date is a good change.

Not sure about using uppercase though, unless you are doing a mass rename to make the APIs more consistent with the types?

S.Option
S.Either

gcanti:
The convention is:
- combinators -> lower case (option, readonlyMap, numberFromString)
- schemas -> lower case if there is a corresponding ts type (number, string), otherwise upper case (DateFromSelf, NumberFromString, Trim)

in `main` the only exception seems `date` at the moment

jessekelly881:
I also hate the pascal case names as well. jaja. Are the full schema definitions really necessary though? `S.NumberFromString` almost seems more awkward to write(and more importantly, read) than `pipe(S.string, S.numberFromString)` jaja

tim-smart:
> The convention is:
> 
> * combinators -> lower case (option, readonlyMap, numberFromString)
> * schemas -> lower case if there is a corresponding ts type (number, string), otherwise upper case (DateFromSelf, NumberFromString, Trim)
> 
> in `main` the only exception seems `date` at the moment

OK I see what you mean now, combinators vs schemas. I think the pascal case is fine then.

src/Schema.ts
```diff
@@ -1528,48 +1532,54 @@ export const date: Schema<Date> = declare(
...
-        typeId: IsValidDateTypeId,
+        typeId: ValidDateTypeId,
         description: "a valid Date",
         ...options
       })
     )
 
 /**
-  This combinator transforms a `string` into a `Date` by parsing the string using `Date.parse`.
+ * A schema representing valid dates, e.g. `new Date("fail")` even if an instance of `Date` is excluded.
```
tim-smart, line 1555: @gcanti I think this reads better:

```
A schema representing valid dates, e.g. `new Date("fail")` is excluded, even though it is an instance of `Date`.
```

PR #238, opened 2023-04-17
chore: update @effect/io

sukovanej:


.changeset/chilly-socks-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+"@effect/schema": patch
```
gcanti, line 2: should be `minor` actually

gcanti (changes requested):

.changeset/chilly-socks-speak.md
sukovanej: fixed

PR #263, opened 2023-04-29
Add fromUnknown

datner:
I started creating fromJson, but I think it requires a more concrete definition


datner:
@gcanti ? 😅

datner:
@gcanti this was part of a conversation about how to express nested vagueness without interlacing `parse` in a `decode`

```ts
interface Foo {
  bar: string,
  data: unknown
}
```

Theres context lost here and a lot of time passed lol.

PR #303, opened 2023-06-17
chore: update @effect/data and fast-check

sukovanej:


pnpm-lock.yaml
```diff
@@ -1,19 +1,19 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
```
gcanti, line 1: FYI if you upgrade to pnpm@latest (8.6.2) this change should vanish

sukovanej: Nice catch, but it's weird, I have `8.6.2` already 🤔

gcanti: Ah, you're right, that's weird. I also tried it with `pnpm@8.6.2` now and I can see the same change... then I have no idea why it was `'6.1'` before...
Well, it doesn't matter, let's merge. Sorry for the noise

PR #310, opened 2023-06-23
chore: update /data and /io

sukovanej:


sukovanej:
Uff `JS heap out of memory`. Just tried with node v16.17.1 locally and all looks good 🤔

PR #318, opened 2023-07-06
Add ulid schema

vecerek:
The PR implements a schema for [ULID](https://github.com/ulid/spec)s, similarly to UUID. This requires an upgrade of `fast-check` where I introduced the ULID arbitrary in https://github.com/dubzzz/fast-check/pull/4020.

### DIFF
--- a/test/data/filter/ULID.ts
+++ b/test/data/filter/ULID.ts
@@ -1,0 +1,18 @@
+import * as S from "@effect/schema/Schema"
+import * as Util from "@effect/schema/test/util"
+
+describe.concurrent("ULID", () => {
+  it("property tests", () => {
+    Util.roundtrip(S.ULID)
+  })
+
+  it("Decoder", async () => {
+    const schema = S.ULID
+    await Util.expectParseSuccess(schema, "01H4PGGGJVN2DKP2K1H7EH996V")
+    await Util.expectParseFailure(
+      schema,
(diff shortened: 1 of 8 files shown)


test/data/filter/ULID.ts
```diff
@@ -0,0 +1,18 @@
...
+    Util.roundtrip(S.ULID)
+  })
+
+  it("Decoder", async () => {
+    const schema = S.ULID
+    await Util.expectParseSuccess(schema, "01H4PGGGJVN2DKP2K1H7EH996V")
+    await Util.expectParseFailure(
+      schema,
+      "",
+      `Expected a string matching the pattern ^[0-7][0-9A-HJKMNP-TV-Z]{25}$, actual ""`
```
gcanti, line 15: FYI if you want a shorter error message, like `Expected ULID, actual ""`, you can add a `title` annotation:

```diff
export const ULID: Schema<string> = pipe(
  string,
  pattern(ulidRegex, {
    typeId: ULIDTypeId,
+   title: "ULID",
    arbitrary: (): Arbitrary<string> => (fc) => fc.ulid()
  })
)
```

vecerek: Thanks for the suggestion! I've force pushed the change.

gcanti:
Thanks @vecerek

PR #331, opened 2023-07-16
chore: update /data and /io

sukovanej:


package.json
```diff
@@ -105,8 +105,8 @@
     "zod": "^3.21.4"
   },
   "dependencies": {
-    "@effect/data": "^0.14.1",
-    "@effect/io": "^0.32.1",
+    "@effect/data": "^0.15.0",
```
gcanti, line 108: @sukovanej `@effect/io@0.32.x` is not compatible with `@effect/data@0.15.0`, I'm waiting for the next release of /io to update /schema

sukovanej: Ah, okay. Sorry, I didn't check.

PR #343, opened 2023-07-19
remove json schema and related types

gcanti:


muradm:
@gcanti, could you please explain why it is removed, and what should one do if used before?

gcanti:
@muradm the reason is to align with `@effect/data` which once had a `Json` module and was then removed. The `Json` type that was once present is not precise and does not satisfy JSON so we prefer not to have it at all.

PR #345, opened 2023-07-20
Fix instanceOf annotations

morvud:
Fixed: Annotations such as `message` are ignored when using `S.instanceOf`.


gcanti:
@morvud thanks!

PR #350, opened 2023-07-23
Add schema for BigintFromString

vecerek:
Implements #351.

Sometimes resource identifiers are very large integers represented as strings, e.g. through query params, and sometimes also via the response body, but stored as integers in the database. This schema will help parse such inputs.

### DIFF
--- a/.changeset/beige-stingrays-move.md
+++ b/.changeset/beige-stingrays-move.md
@@ -1,0 +1,5 @@
+---
+"@effect/schema": patch
+---
+
+Add schema for BigintFromString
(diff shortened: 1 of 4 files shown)


.changeset/beige-stingrays-move.md
```diff
@@ -0,0 +1,5 @@
+---
+"@effect/schema": minor
```
gcanti, line 2: This should be "patch" (before v1: minor=breaking change, patch=otherwise)

gcanti:
Thanks @vecerek, could you please also add some documentation to the README? Something along the lines of numberFromString is fine

gcanti (approved):

PR #357, opened 2023-07-24
feat: add `isSchema` guard

sukovanej:
In effect-http, I need a runtime check for whether an object is a `Schema`. I could do it in the lib structurally by checking the `ast` field but that is verbose and probably prone to errors when the AST changes.

### DIFF
--- a/src/Schema.ts
+++ b/src/Schema.ts
@@ -19,7 +19,7 @@ import * as O from "@effect/data/Option"
 import type { Pipeable } from "@effect/data/Pipeable"
 import { pipeArguments } from "@effect/data/Pipeable"
 import type { Predicate, Refinement } from "@effect/data/Predicate"
-import { isDate } from "@effect/data/Predicate"
+import { isDate, isObject } from "@effect/data/Predicate"
 import * as RA from "@effect/data/ReadonlyArray"
 import type { Arbitrary } from "@effect/schema/Arbitrary"
 import type { ParseOptions } from "@effect/schema/AST"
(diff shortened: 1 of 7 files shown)


src/Schema.ts
```diff
@@ -224,6 +224,15 @@ class SchemaImpl<From, To> implements Schema<From, To> {
...
 export const make = <I, A>(ast: AST.AST): Schema<I, A> => new SchemaImpl(ast)
 
+/**
+ * Tests if a value is a `Schema`.
+ *
+ * @category guards
+ * @since 1.0.0
+ */
+export const isSchema = (input: unknown): input is Schema<unknown, unknown> =>
+  input instanceof SchemaImpl
```
gcanti, line 234: I think we should follow the convention in `@effect/data` and add a `readonly _id: TypeId` prop, see for example https://github.com/Effect-TS/data/blob/main/src/Duration.ts

gcanti (changes requested):

src/Schema.ts
sukovanej: Yop, done.

test/Schema.ts
```diff
@@ -360,4 +360,9 @@ describe.concurrent("Schema", () => {
       new Error("`optionalElement` is not supported on this schema")
     )
   })
+
+  it("isSchema", () => {
+    expect(S.isSchema(S.string)).toBe(true)
+    expect(S.isSchema(S.struct({ f: S.optional(S.string).toOption() }))).toBe(true)
+  })
```
gcanti, line 367: `isSchema` should account for `BrandSchema`:

```ts
expect(S.isSchema(S.number.pipe(S.int(), S.brand("Int")))).toBe(true)
```

but we need this fix https://github.com/Effect-TS/data/pull/438 first since `isObject` mistakingly doesn't validate functions

gcanti (changes requested):

test/Schema.ts
gcanti: @sukovanej patch released, could you please upgrade to latest `@effect/data` and add the BrandSchema test (which should fail). If I'm not mistaken the fix should be just a matter of adding `_id: TypeId` to `out` here https://github.com/Effect-TS/schema/blob/2729d7ace3c70c88e148bb025ca6c6a2a61b92f3/src/Schema.ts#L773-L786

sukovanej: Exactly as you said, thanks a lot! Done.

Issue #352, opened 2023-07-23
Add schema for parsing arrays from delimited strings


vecerek:
## 🚀 Feature request

A function that returns a schema that can parse delimited strings into arrays of strings.

### Current Behavior

Missing

### Desired Behavior

```ts
const commaSeparatedList = S.arrayFromDelimitedString(",")
const parse = S.parseSync(commaSeparatedList)

const a: readonly string[] = parse("1,2,3")
// [1,2,3]
```

### Who does this impact? Who is this for?

This can be helpful to both HTTP clients and servers that have a need to parse query params representing a list of ids or other values, e.g.:

```
GET /api/users/show_many?ids=123,456,789
```

### Your environment

| Software      | Version(s) |
| ------------- | ---------- |
| @effect/schema |            |
| TypeScript    |            |

vecerek:
Now available since https://github.com/Effect-TS/schema/releases/tag/v0.30.4.

The way to achieve is different from the proposal, though:

```ts
//$ExpectType Schema<string, ReadonlyArray<string>>
S.string.pipe(S.split(","))

//$ExpectType Schema<string, ReadonlyArray<number>>
S.string.pipe(S.split(","), S.compose(S.array(S.NumberFromString)))
```

PR #361, opened 2023-07-26
Add `compose`

vecerek:
Fully addresses https://github.com/Effect-TS/schema/issues/352 in combination with https://github.com/Effect-TS/schema/pull/359 as opposed to the approach outlined in https://github.com/Effect-TS/schema/pull/353.

Based on the suggestion from https://github.com/Effect-TS/schema/pull/359#issuecomment-1651019819.

The `compose` combinator allows one to directly transform `Schema<A, B>` to `Schema<B, C>` creating a `Schema<A, C>`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -983,6 +983,15 @@ S.struct({ a: S.string, b: S.string }).pipe(
 );
 ```
 
+## Compose
+
+The `compose` combinator allows you to combine two schemas.
+
+```ts
+// $ExpectType Schema<string, readonly number[]>
+S.compose(S.split(S.string, ","), S.array(S.NumberFromString));
+```
+
 ## InstanceOf
 
 In the following section, we demonstrate how to use the `instanceOf` combinator to create a `Schema` for a class instance.
--- a/.changeset/nice-ants-push.md
+++ b/.changeset/nice-ants-push.md
@@ -1,0 +1,5 @@
+---
+"@effect/schema": patch
+---
+
+Adds `compose`: a combinator that composes Schema<A, B> with Schema<B, C> into Schema<A, C>.
--- a/src/Schema.ts
+++ b/src/Schema.ts
@@ -917,6 +917,19 @@ export const extend: {
  * @category combinators
  * @since 1.0.0
  */
+export const compose: {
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -983,6 +983,26 @@ S.struct({ a: S.string, b: S.string }).pipe(
 );
 ```
 
+## Compose
+
+The `compose` combinator allows you to directly transform one schema to another. It is a convenience function over [transform](#transform) with two identity functions.
```
gcanti, line 988: > transform one schema to another

`compose` combines **two** schemas

> It is a convenience function over [transform](#transform) with two identity functions.

I wouldn't leak the implementation (in the future we could change it or we could even have a primitive in AST)

README.md
```diff
@@ -1239,29 +1259,6 @@ parse(3); // 1
 
 ### Bigint transformations
 
-#### BigintFromString
```
gcanti, line 1242: removing this documentation is an unwanted side effect I guess

test/data/ReadonlyArray.ts
```diff
@@ -49,3 +49,5 @@ describe.concurrent("ReadonlyArray", () => {
     )
   })
 })
+
+S.string.pipe(S.split(","), S.compose(S.array(S.NumberFromString)));
```
gcanti, line 53: not sure I understand the purpose of this

gcanti (changes requested):

README.md
vecerek: oops, absolutely!

test/data/ReadonlyArray.ts
vecerek: Ahh, I was scribbling something out and forgot to remove it 🤦

README.md
vecerek: I've no idea how these deletions came to be. Probably a mistake that slipped in during the rebase.

README.md
vecerek: Addressed in https://github.com/Effect-TS/schema/pull/361/commits/59154dae1a147bc7ae642d03bd7f3dc7d313c120

gcanti (approved):

PR #362, opened 2023-07-26
Make `compose` more flexible to work with

vecerek:
Implements #369.

Previously the following schema would not type check:

```ts
const schema = S.record(S.string, S.unknown).pipe(S.compose(S.struct({ a: S.string })))
```

With this change, `compose` becomes more flexible in a sense that it allows the composition of `Schema<A, B1>` and `Schema<B2, C>` into `Schema<A, C>` if and only if `B2 extends B1`.

### DIFF
--- a/.changeset/proud-candles-itch.md
+++ b/.changeset/proud-candles-itch.md
@@ -1,0 +1,5 @@
+---
+"@effect/schema": patch
+---
+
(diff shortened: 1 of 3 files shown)


gcanti:
Closed by https://github.com/Effect-TS/schema/pull/388

PR #365, opened 2023-07-27
make split a bit smarter

datner:
- test(dtslint): add string split test
- chore: add missing separator
- feat(Schema): allow smarter split type behavior
- test(dtslint): add split string literal test
- chore(nix): add corepack support

### DIFF
--- a/.changeset/cuddly-pandas-act.md
+++ b/.changeset/cuddly-pandas-act.md
@@ -1,0 +1,5 @@
+---
+"@effect/schema": patch
+---
+
+make split smarter
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ tmp/
 build/
 dist/
 .direnv/
+out/
--- a/dtslint/Schema.ts
+++ b/dtslint/Schema.ts
@@ -102,6 +102,9 @@ pipe(S.string, S.endsWith('a'));
 // $ExpectType Schema<string, string>
 pipe(S.string, S.includes('a'));
 
+// $ExpectType Schema<string, ReadonlyArray<string>>
+pipe(S.string, S.split('.'));
+
 // $ExpectType Schema<number, number>
 pipe(S.number, S.greaterThan(5));
 
--- a/dtslint/Schema.ts
+++ b/dtslint/Schema.ts
@@ -105,6 +105,10 @@ pipe(S.string, S.includes('a'));
 // $ExpectType Schema<string, ReadonlyArray<string>>
 pipe(S.string, S.split('.'));
 
+// ---------------------------------------------
+// numbers
+// ---------------------------------------------
+
 // $ExpectType Schema<number, number>
 pipe(S.number, S.greaterThan(5));
 
--- a/dtslint/Schema.ts
+++ b/dtslint/Schema.ts
@@ -483,6 +483,9 @@ const FooterLocaleIDs = S.literal("footer_title", "footer_sendoff")
 // $ExpectType Schema<"welcome_email_id" | "email_heading_id" | "footer_title_id" | "footer_sendoff_id", "welcome_email_id" | "email_heading_id" | "footer_title_id" | "footer_sendoff_id">
 S.templateLiteral(S.union(EmailLocaleIDs, FooterLocaleIDs), S.literal("_id"))
 
+// $ExpectType Schema<`a.${string}`, readonly ["a", ...string[]]>
+pipe(S.templateLiteral(S.literal('a'), S.literal('.'), S.string), S.split("."))
+
 // ---------------------------------------------
 // attachPropertySignature
 // ---------------------------------------------
--- a/flake.lock
+++ b/flake.lock
@@ -20,11 +20,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1690083312,
-        "narHash": "sha256-I3egwgNXavad1eIjWu1kYyi0u73di/sMmlnQIuzQASk=",
+        "lastModified": 1690593349,
+        "narHash": "sha256-i6jdORO+YiP19pFNeR7oYIIwmzQvdxwNO+BmtATcYpA=",
         "owner": "nixos",
         "repo": "nixpkgs",
-        "rev": "af8cd5ded7735ca1df1a1174864daab75feeb64a",
+        "rev": "11cf5e1c74fe6892e860afeeaf3bfb84fdb7b1c3",
         "type": "github"
       },
       "original": {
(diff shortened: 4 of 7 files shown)


datner:
@gcanti review me please <3

gcanti:
I can review the code for `Schema.ts`, but I'll leave the changes to `flake.nix` to @IMax153

IMax153:
@datner why do we need core pack? We do not have it enabled in any of our other repositories and we really try to not introduce changes in the build process unless absolutely necessary. 

Could you outline the benefits?

datner:
@IMax153 if you install pnpm from nix, you get the specific version that nix has. Which is 7.11.21 or something. But not all the users use nix, which creates conflicts and noise around the pnpm-lock.

if you use corepack and the `packageManager` field in package json, all users of corepack and nix will use the correct versions. I think that those who are invariant also get a warning for it if they use  a non compatible version.

I've segmented everything to commits so it's clear which changes relate to which. take a look

tl;dr
it makes sure that both nix users and @gcanti use the same pnpm version 
(and the occasion contributor ofc)

IMax153:
@datner I think this makes a lot of sense. However, your implementation is slightly incorrect. We can enable `corepack` and add it as a package which prevents us from having to use the `shellHook` or manipulate the `PATH` manually.

```nix
{
  inputs = {
    nixpkgs = {
      url = "github:nixos/nixpkgs/nixpkgs-unstable";
    };

    flake-utils = {
      url = "github:numtide/flake-utils";
    };
  };

  outputs = {
    self,
    nixpkgs,
    flake-utils,
    ...
  }:
    flake-utils.lib.eachDefaultSystem (system: let
      pkgs = nixpkgs.legacyPackages.${system};
      corepackEnable = pkgs.runCommand "corepack-enable" {} ''
        mkdir -p $out/bin
        ${pkgs.nodejs-18_x}/bin/corepack enable --install-directory $out/bin
      '';
    in {
      formatter = pkgs.alejandra;

      devShells = {
        default = pkgs.mkShell {
          buildInputs = with pkgs; [
            nodejs-18_x
            corepackEnable
          ];
        };
      };
    });
}
```

vecerek (commented):
👍 for the changes to `split`, very similar to my proposed changes to `compose` in #362

IMax153:
@gcanti - this is now good to go from a Nix perspective.

jessekelly881:
S.split could decode to a nonempty array, no?

vecerek:
@datner I think his answer will be similar to
> @vecerek yeah it seems handy, I need a little more time to think about all the consequences of this change though (in particular whether it will play well with schemas being invariant)

([link](https://github.com/Effect-TS/schema/issues/369#issuecomment-1658664036))

I think it's a worthy design goal to preserve the roundtrip property of schemas. But then, we also need a way to break the invariance for one-way transformations, e.g. when we only care about decoding.

gcanti:
@vecerek yeah, the two changes are kind of connected. This is the current signature of `split`:

```ts
declare const split: {
  (separator: string): <I>(self: S.Schema<I, string>) => S.Schema<I, readonly string[]>
  <I>(self: S.Schema<I, string>, separator: string): S.Schema<I, readonly string[]>
}
```

Note that it could be "smarter", i.e. the return type could be `NonEmptyReadonlyArray<string>` since `@effect/data/String#split` returns a `NonEmptyArray<string>`

However it would lead to issues when you try to compose (invariance kicks in)

```ts
import * as S from '@effect/schema/Schema'
import { NonEmptyReadonlyArray } from 'effect/ReadonlyArray'

declare const smarterSplit: {
  (separator: string): <I>(self: S.Schema<I, string>) => S.Schema<I, NonEmptyReadonlyArray<string>>
  <I>(self: S.Schema<I, string>, separator: string): S.Schema<I, NonEmptyReadonlyArray<string>>
}

S.compose(S.split(S.string, '.'), S.array(S.NumberFromString)) // OK

S.compose(smarterSplit(S.string, '.'), S.array(S.NumberFromString)) // ERROR
/*
Argument of type 'Schema<string, readonly [string, ...string[]]>' is not assignable to parameter of type 'Schema<string, readonly string[]>'.
  Types of property 'To' are incompatible.
    Type '(_: readonly [string, ...string[]]) => readonly [string, ...string[]]' is not assignable to type '(_: readonly string[]) => readonly string[]'.
      Types of parameters '_' and '_' are incompatible.
        Type 'readonly string[]' is not assignable to type 'readonly [string, ...string[]]'.ts(2345)
*/
```

datner:
> is it really worth trying to be so precise with the return type at the risk of encountering uncovered corner cases? What is the use case?

@gcanti well, there are 2 reasons why you would use split
1. you got a list encoded as a delimited string `"foo,bar,baz,qux,quux"`
2. you have some structure encoded as a delimited string `"scope:name:datatype:value"` (<- I have a similar structure in my work. Not defined by me, defined by Microsoft 🙃 )

the first case is ofc covered by the regular split, the latter would swallow my types and force me to re-check them
```ts
const oof = S.templateLiteral(/* structure from point 2 */).pipe( // Schema<string, `${string}:${string}:${string}:${string}">
  S.split(":") // <- now it's back to string[],
  S.checkAgainThatItsStillTheStructureFromPoint2ButNowAsTuple
 )
```

> Note that it could be "smarter", i.e. the return type could be NonEmptyReadonlyArray<string> since @effect/data/String#split returns a NonEmptyArray<string>
However it would lead to issues when you try to compose (invariance kicks in)

Ok, so what do we do about that? 😅

mikearnaldi:
> @datner I think his answer may be similar to
> 
> > @vecerek yeah it seems handy, I need a little more time to think about all the consequences of this change though (in particular whether it will play well with schemas being invariant)
> 
> ([link](https://github.com/Effect-TS/schema/issues/369#issuecomment-1658664036))
> 
> I think it's a worthy design goal to preserve the roundtrip property of schemas. But then, we also need a way to break the invariance for one-way transformations, e.g. when we only care about decoding. I wonder if we could come up with a method that would return a schema that can only decode but loses the encoding capability on the type level.
> 
> UPDATE: After a bit more thought, I'm actually not sure if I understood Giulio properly. The roundtrip property is still preserved because of the "composed" type, e.g. the template literal in your case. So maybe the invariance is about something else 🤷

This misses the point of Schema, invariance isn't related to encoding / decoding. `Schema<I, A>` contains a description of `I` and a description of `A`  given a schema you can derive an arbitrary `F<A>` so it is totally valid for example to derive `Equivalence<A>` and `Decoder<A>`. In Equivalence the type A is an input, in Decoder it is an output. So in schema invariance is a must, because of what schema describes.

PR #372, opened 2023-08-03
add `Class` to /Schema module

fubhy:
@mikearnaldi Not sure if this is what you meant... 

If it is, please give @tim-smart the commit credit for this (set `--author` of the commit) because it was literally just a matter of copy & pasting his code over ;-)


docs/modules/SchemaClass.ts.md
```diff
@@ -0,0 +1,254 @@
...
+## SchemaClass overview
+
+Added in v1.0.0
+
+---
+
+<h2 class="text-delta">Table of contents</h2>
+
+- [constructor](#constructor)
+  - [SchemaClass](#schemaclass)
```
tim-smart, line 16: We should rename this to `Base`, and call the others `Extends`, `Transform` etc

So usage will be:

```ts
import * as SchemaClass from "@effect/schema/SchemaClass"

class User extends SchemaClass.Base({ ... }) {}
class UserWithAge extends SchemaClass.Extends(User, { age: Schema.number }) {}
```

docs/modules/SchemaClass.ts.md
```diff
@@ -0,0 +1,254 @@
...
+  - [SchemaClassExtends](#schemaclassextends)
+  - [SchemaClassTransform](#schemaclasstransform)
+  - [SchemaClassTransformFrom](#schemaclasstransformfrom)
+- [model](#model)
+  - [CopyWith (interface)](#copywith-interface)
```
tim-smart, line 21: We don't need to export this.

src/SchemaClass.ts
```diff
@@ -0,0 +1,396 @@
...
+import type { ParseError, ParseResult } from "@effect/schema/ParseResult"
+import type {
+  From,
+  FromOptionalKeys,
+  PropertySignature,
+  Schema,
+  Spread,
+  To,
+  ToOptionalKeys
+} from "@effect/schema/Schema"
```
tim-smart, line 15: Lets change this to a namespace import.

```ts
import * as S from "@effect/schema/Schema"
```

It is more consistent with the rest of the codebase.

src/SchemaClass.ts
```diff
@@ -0,0 +1,396 @@
...
+import {
+  from,
+  instanceOf,
+  struct,
+  to,
+  transform,
+  transformResult,
+  validate,
+  validateSync
+} from "@effect/schema/Schema"
```
tim-smart, line 25: Same as above.

src/SchemaClass.ts
```diff
@@ -0,0 +1,396 @@
...
+  fn.unsafe = function unsafe(props: unknown) {
+    return Object.assign(Object.create(this.prototype), props)
+  }
+  fn.structSchema = function structSchema() {
+    return schema_
+  }
+  fn.schema = function schema(this: any) {
+    return transform(
+      schema_,
+      instanceOf(this),
```
tim-smart, line 157: Replace this with `Schema.unknown` or `Schema.any`. Whatever compiles without casting.

src/SchemaClass.ts
```diff
@@ -0,0 +1,396 @@
...
+    )
+  }
+  fn.prototype.copy = function copy(this: any, props: any) {
+    return new (this.constructor as any)({
+      ...this,
+      ...props
+    })
+  }
+  fn.prototype.unsafeCopy = function unsafeCopy(this: any, props: any) {
+    return Object.setPrototypeOf(
```
tim-smart, line 169: I think this should be:

```ts
Object.assign(Object.create(this.constructor.prototype), this, props)
```

tim-smart (changes requested):

docs/modules/SchemaClass.ts.md
tim-smart: OK scratch that. Lets add this to the /Schema module as `Class` (as per the Discord discussion).

```ts
import * as Schema from "@effect/schema/Schema"

class User extends Schema.Class({ ... }) {}
class UserWithAge extends Schema.ClassExtends(User, { age: Schema.number }) {}
```

docs/modules/Schema.ts.md
```diff
@@ -32,7 +32,7 @@ Added in v1.0.0
 - [bigint](#bigint)
   - [BigintFromString](#bigintfromstring)
   - [betweenBigint](#betweenbigint)
-  - [bigintFromString](#bigintfromstring)
+  - [bigintFromString](#bigintfromstring-1)
```
fubhy, line 35: Unrelated but noticed that these are duplicated currently. Adding a numerical suffix causes the links to jump to the right spot.

... Can do that in a separate PR of course.

src/Schema.ts
```diff
@@ -1183,6 +1184,375 @@ export const documentation =
   (documentation: AST.DocumentationAnnotation) => <I, A>(self: Schema<I, A>): Schema<I, A> =>
     make(AST.setAnnotation(self.ast, AST.DocumentationAnnotationId, documentation))
 
+// ---------------------------------------------
+// class
```
fubhy, line 1188: I'll split this up correctly.

test/SchemaClass.ts
fubhy, line 1: I left the tests in a separate file for now.

docs/modules/Schema.ts.md
tim-smart: The docs are automatically generated, so we will need to patch our tooling. (cc @gcanti )

src/Schema.ts
tim-smart: I think a seperate class section is a good idea :)

src/Schema.ts
```diff
@@ -1183,6 +1184,375 @@ export const documentation =
...
+// ---------------------------------------------
+
+/**
```
tim-smart, line 1191: You can remove this comment as it isn't exported.

src/Schema.ts
```diff
@@ -1183,6 +1184,375 @@ export const documentation =
...
+ * @category model
+ * @since 1.0.0
+ */
+interface CopyWith<A> {
+  copy<T>(this: T, props: Partial<A>): T
+  unsafeCopy<T>(this: T, props: Partial<A>): T
+}
+
+/**
+ * @category model
```
tim-smart, line 1201: I think just use a doc category of "classes" for all of these exports.

test/SchemaClass.ts
tim-smart: I think we can rename it to `test/classes.ts`

tim-smart (changes requested):
Thanks, almost there I think! @gcanti did you want to take a look?

tim-smart (approved):

gcanti:
It would be helpful to have some documentation on how to use it for end users

src/Schema.ts
```diff
@@ -1183,6 +1184,371 @@ export const documentation =
...
+    props: A
+  ): InstanceType<T>
+
+  schema<T extends new(...args: any) => any>(
+    this: T
+  ): Schema<I, InstanceType<T>>
+
+  structSchema(): Schema<I, A>
+
+  readonly fields: Record<string, Schema<I, A>>
```
gcanti, line 1219: The type `Record<string, Schema<I, A>>` doesn't seem correct:

```ts
class A extends S.Class({
  a: S.string,
  b: S.number
}) {}
const fields = A.fields
/*
const fields: Record<string, S.Schema<{
  readonly a: string;
  readonly b: number;
}, {
  readonly a: string;
  readonly b: number;
}>>
*/
```

Maybe should be `/** @internal */`, what's the purpose of a public `fields`?

src/Schema.ts
```diff
@@ -1183,6 +1184,371 @@ export const documentation =
   (documentation: AST.DocumentationAnnotation) => <I, A>(self: Schema<I, A>): Schema<I, A> =>
     make(AST.setAnnotation(self.ast, AST.DocumentationAnnotationId, documentation))
 
+// ---------------------------------------------
+// classes
+// ---------------------------------------------
+
+interface CopyWith<A> {
+  copy<T>(this: T, props: Partial<A>): T
```
gcanti, line 1192: `copy` and `unsafeCopy` are reserved but there's no alert for the user

```ts
it('', () => {
  class A extends S.Class({
    copy: S.string
  }) {}
  const a = new A({ copy: '-' })
  console.log(a.copy({})) // throws
})
```

maybe we could move `copy` and `unsafeCopy` to static functions on the class to avoid name clashes in the first place

gcanti (changes requested):

src/Schema.ts
fubhy: Nothing comes to mind off the top of my head and I generally prefer to keep things private until a use case emerges to keep the public interface narrow to leave wiggle room for maintenance & change. Would you agree @tim-smart ?

fubhy: Also just noticed that `fields` is currently not exposed on `ClassTransform`. If we want to keep it we should add it there too.

fubhy: The fact that it's also not  exposed on `ClassTransform` is imho an indicator that this is indeed internal.

fubhy: If we were to keep it public I'd type it sth. like this:

```typescript
type Fields<A, B> = keyof A extends keyof B ? keyof B extends keyof A ? {
      readonly [K in keyof A & keyof B]: Schema<A[K], B[K]>
    } :
  never :
  never
```

tim-smart: Yes you can make that private.

README.md
```diff
@@ -910,6 +910,39 @@ encode({ a: O.none() }) // {}
...
+  }
+}
+
+const person = new Person({ id: 1, name: "Tim" }); // constructors validate the props
+const parsePerson = S.parse(Person.schema());
+
+assert(person instanceof Data.Class); // extends Data for equality checks
+
+// clone a instance and validate the props
+const john = person.copyWith({ name: "John" });
```
tim-smart, line 941: Will need to update this to use the static method.

README.md
```diff
@@ -910,6 +910,39 @@ encode({ a: O.none() }) // {}
...
+
+assert(john.id === 1);
```
tim-smart, line 943: I feel like this can be removed, it doesn't really illustrate much.

tim-smart (changes requested):

README.md
```diff
@@ -910,6 +910,37 @@ encode({ a: O.none() }) // {}
...
+  }
+}
+
+const tim = new Person({ id: 1, name: "Tim" }); // constructors validate the props
+const parsePerson = S.parse(Person.schema());
+
+assert(person instanceof Data.Class); // extends Data for equality checks
+
+// clone a instance and validate the props
+const john = Person.copyWith(tim, { name: "John" });
```
tim-smart, line 941: It's `Person.copy()` now :)

tim-smart (changes requested):

README.md
fubhy: Aw maaaaaaan, I'm sorry... I'm playing Baldurs Gate atm. and not paying attention properly -.-

tim-smart: Lol

tim-smart (approved):

docs/modules/Schema.ts.md
tim-smart: Updated docgen to fix this.

gcanti:
RE `copy` / `unsafeCopy`: looks like there's no way to remove an optional field:

```ts
class A extends S.Class({
  a: S.optional(S.string),
  b: S.number
}){}
const a1 = new A({ a: 'a', b: 1 })

// I can update `b` but how can I remove the `a` field?
const a2 = A.copy(a1, { b: 2 })
```

src/SchemaClass.ts
gcanti: `unknown` doesn't seem correct (it will `S.validate` any value), must be a schema representing `Schema.Class<I, A>`:

```ts
it("schema", () => {
  // existing test
  const person = S.parseSync(Person.schema())({ id: 1, name: "John" })
  assert(person.name === "John")
  // added
  const PersonFromSelf = S.to(Person.schema())
  Util.expectParseSuccess(PersonFromSelf, new Person({ id: 1, name: "John" }))
  Util.expectParseFailure(PersonFromSelf, { id: 1, name: "John" }, `Expected an instance of fn, actual {"id":1,"name":"John"}`)
})
```

fubhy:
> RE `copy` / `unsafeCopy`: looks like there's no way to remove an optional field:
> 
> ```ts
> class A extends S.Class({
>   a: S.optional(S.string),
>   b: S.number
> }){}
> const a1 = new A({ a: 'a', b: 1 })
> 
> // I can update `b` but how can I remove the `a` field?
> const a2 = A.copy(a1, { b: 2 })
> ```

Good shout. There are three options that I can think of right now.

1. Accept that this is a limitation of `.copy`
2. Break the `exactOptionalProperties` contract and accept `undefined` as "delete".
3. Introduce a special symbol that is accepted in place of to-be-deleted fields, e.g:

```ts
A.copy(a, { b: S.Class.Delete })
```

The second option would likely not be a good path forward. I personally think that both 1. and 3. are acceptable. I think I lean slightly towards option 3.

gcanti:
I vote for 2, as it is an exception: no other schema has a `copy` function, and the goals of `Class` are (from the README):

- Define a schema and an opaque type in one pass
- Attach common functionality using class methods or getters
- Implement Data.Case for checking equality by value and hashing

Alternatively we could try to also give all other schemas this functionality, but it may be better to postpone this to another PR and focus on the main goals of `Class` for now.

---

Other 2 observations:

1) Another function that I find confusing is `effect`. At first glance, it is not clear what it does (`parse`, `decode`, `validate`?), as the name describes the result rather than the operation (btw the code shows that it uses `P.validate`).

    I guess that the `effect` operation is strictly necessary when the schemas involved in the fields contain, for example, async transformations and then calling `new Person()` to construct a value would always throw an error. However, the documentation does not explicitly state this and makes it seem like the operation is kind of an additional feature.

2) Sorry if it has been already discussed, but what about

    ```ts
    class PersonWithAge extends Person.extend({
      age: S.number
    }) {
      get isAdult() {
        return this.age >= 18;
      }
    }
    ```

    instead of

    ```ts
    class PersonWithAge extends S.ClassExtends(Person, {
      age: S.number
    }) {
      get isAdult() {
        return this.age >= 18;
      }
    }
    ```

fubhy:
> I vote for 2

:+1: 

> 1. Another function that I find confusing is effect

@tim-smart Is likely better equipped to defend that one or propose a different name. 

> 2. Sorry if it has been already discussed, but what about

This makes sense imho and is more in-line with the composability theme of the package in general.

--- 

While we are at it... I'd probably challenge the existence of `.unsafe` too. That looks like a footgun to me.

gcanti:
> While we are at it... I'd probably challenge the existence of .unsafe too. That looks like a footgun to me.

I agree, I can see the practical benefit of a constructor that may throw on invalid field values, but what's the use case of `unsafe`?

Another nitpick: there's a lot of duplication here 

https://github.com/Effect-TS/schema/pull/372/files#diff-54dec539aad7e644bf1e312230fb30da27d9ebc7c17e055ae7ec229383d5b8c4R1198-R1224

all these functions are used by `Class`, `ClassExtends` and `ClassTransform` (well almost, looks like `ClassTransform` doesn't have `effect`, is that intentional?), we could move these functions to a base interface.

fubhy:
> all these functions are used by Class, ClassExtends and ClassTransform (well almost, looks like ClassTransform doesn't have effect, is that intentional?), we could move these functions to a base interface.

How'd you feel about also adding some type aliases for the shared types between `struct` and this?

E.g. sth. like this:

```ts
type StructFields = Record<
  PropertyKey,
  | Schema<any>
  | Schema<never>
  | PropertySignature<any, boolean, any, boolean>
  | PropertySignature<never, boolean, never, boolean>
>

type StructFrom<Fields extends StructFields> =
  & { readonly [K in Exclude<keyof Fields, FromOptionalKeys<Fields>>]: From<Fields[K]> }
  & { readonly [K in FromOptionalKeys<Fields>]?: From<Fields[K]> }

type StructTo<Fields extends StructFields> =
  & { readonly [K in Exclude<keyof Fields, ToOptionalKeys<Fields>>]: To<Fields[K]> }
  & { readonly [K in ToOptionalKeys<Fields>]?: To<Fields[K]> }
```

I generally prefer aliasing types whenever they start to get too long & complex to have them in-line in a function signature. By giving them an expressive name the function signature (at least to me) is more readable.

fubhy:
@gcanti I believe I addressed your feedback except for your suggestion about `Person.extend(...)` and your comment on the ambiguity of `.effect(...)`. 

WDYT there @tim-smart ?

tim-smart:
Yes we can remove `effect`, as you can just use Schema.validate if required.

Adding `.extend` and `.transform` seems like a good change. Did you want to attempt it?

tim-smart:
I'll ask a couple people who are using `effect-schema-class` in large apps, what features they are using.

fubhy:
> Yes we can remove `effect`, as you can just use Schema.validate if required.
> 
> Adding `.extend` and `.transform` seems like a good change. Did you want to attempt it?

Yeah I can tackle that in the morning!

tim-smart:
As for copy and removing fields, I think you would need to add a separate API for removing optional fields.

fubhy:
> As for copy and removing fields, I think you would need to add a separate API for removing optional fields.

Yeah but then how far do we take that? What about nested structs and optional fields there?

tim-smart:
> Yeah but then how far do we take that? What about nested structs and optional fields there?

FWIW I'm OK with the removal, as you can easily add your own copy methods.

If we do add something it would only be for shallow copies - deep copies would add too much complexity for a feature with low usage (at least I don't deep copy very often).

And the instance method API is more preferable. Even something like:

```ts
person.$copy({ age: 18 }).$remove("address")
```

fubhy:
I took the liberty to consolidate and (imho) simplify the interface a bit.

gcanti:
What's the use case of `transform` and `transformFrom`?

tim-smart:
> What's the use case of `transform` and `transformFrom`?

If you need to do some transformations (mainly effectful, like embellishing a entity from a database etc) and then making a class from the result.

It is a rare use case, but does come in useful.

tim-smart (commented):
Just fixed up inheritance, and removed fields as a static property.

Not sure about the naming of `.struct()`. `.schemaStruct()` might be better as it will autocomplete alongside `.schema()`.

fubhy:
Thanks @tim-smart. I think we are close now!

> Not sure about the naming of `.struct()`. `.schemaStruct()` might be better as it will autocomplete alongside `.schema()`.
I don't feel strongly about this.  Happy to rename it again.

gcanti:
> If you need to do some transformations (mainly effectful, like embellishing an entity from a database etc) and then making a class from the result.

I see, thanks. It is not immediately clear, I think it would be helpful to add a realistic example of both functions to the documentation (which is currently missing).

> .schemaStruct() might be better as it will autocomplete alongside .schema()

👍

src/Schema.ts
```diff
@@ -1183,6 +1196,150 @@ export const documentation =
...
+
+/**
+ * @category classes
+ * @since 1.0.0
+ */
+export interface Class<I, A, Inherited = {}> {
+  new(props: A): A & D.Case & Omit<Inherited, keyof A>
+
+  schema<T extends new(...args: any) => any>(this: T): Schema<I, InstanceType<T>>
+  structSchema(): Schema<I, A>
```
gcanti, line 1211: `structSchema` ->`schemaStruct`

fubhy: Ah. It was `structSchema` before. But agreed. Reads better together with `.schema` if it's `schemaStruct`.

gcanti (approved):

gcanti:
Thank you @fubhy

ljani:
> What's the use case of transform and transformFrom?

I'd like to do some data mapping from wire format to domain model. I'm not sure if the following example is possible at the moment as it seems the original fields must be present in the other representation?

```ts
export const PingRequest = pipe(
    S.struct({
        t: S.literal("p"),
        d: S.record(S.never, S.never),
    }),
    S.transform(
        S.struct({
            type: S.literal("ping"),
        }),
        () => ({ type: "ping" as const }),
        () => ({ t: "p" as const, d: {} })
    )
);
```

I'm more complex scenarios I'm composing types in the `d`/`data` field and it seems to work fine. Though it was easier to compose with `fp-ts`, because you could directly pass the schema instead of having to explicitly decode/encode it, but it isn't a big deal:

```ts
export const ControlRequest = pipe(
    S.struct({
        t: S.literal("c"),
        d: ControlRequestPayload,
    }),
    S.transformResult(
        S.struct({
            type: S.literal("control"),
            data: ControlRequestPayload,
        }),
        ({ d }) =>
            pipe(
                d,
                S.encodeEither(ControlRequestPayload),
                E.mapRight((data) => ({ type: "control" as const, data }))
            ),
        ({ data }) =>
            pipe(
                data,
                S.decodeEither(ControlRequestPayload),
                E.mapRight((d) => ({ t: "c" as const, d }))
            )
    )
);
```

PR #377, opened 2023-08-04
Schema.chunk fails with empty chunks

fubhy:


test/data/Chunk.ts
```diff
@@ -74,6 +74,7 @@ describe.concurrent("Chunk", () => {
   it("chunk. decoder", async () => {
     const schema = S.chunk(S.number)
     await Util.expectParseSuccess(schema, [], C.empty())
+    await Util.expectParseSuccess(schema, C.empty(), C.empty())
```
gcanti, line 77: @fubhy `S.chunk` is not supposed to work with `Chunk`s, it works with arrays as input (while parsing / decoding)

fubhy: I see. I also found `chunkFromSelf` is that what is supposed to be used here?

I find it slightly confusing that for other similar cases (e.g. String => Number coercion) we've got dedicated express utils.

I expected `S.chunk` to validate that the input is a chunk, not to coerce a chunk from an array.

gcanti: The rationale, is to use the most intuitive name for the default codec, for example:

- number: the default codec does nothing (a number is already encodable to JSON), so it is called `Schema.number`. Parsing a number from a string is an exception, not the rule, so it is called `Schema.NumberFromString`.
- chunk: the default codec should allow decoding and encoding to JSON, so it is called `Schema.chunk`. Validating / transforming a chunk is an exception, in the sense that it is less common, not the rule, so it is called `Schema.chunkFromSelf`.

fubhy: Ok, that makes sense. Thanks for the explanation of the rationale!

PR #382, opened 2023-08-06
add `bigIntFromNumber` transform

fubhy:
I find myself needing this quite often. Might be worth it to include it here?


gcanti:
> Looks like it doesn't account for encoding errors

@fubhy should be fixed https://github.com/Effect-TS/schema/pull/393

fubhy:
> > Looks like it doesn't account for encoding errors
> 
> @fubhy should be fixed #393

Works! Thanks.

src/Schema.ts
```diff
@@ -1705,6 +1705,48 @@ export const bigintFromString = <I, A extends string>(self: Schema<I, A>): Schem
...
+    bigint,
+    (n) => {
+      try {
+        return PR.success(BigInt(n))
+      } catch (_) {
+        return PR.failure(PR.type(schema.ast, n))
+      }
+    },
+    (b) => {
+      if (b > BigInt(Number.MAX_SAFE_INTEGER) || b < BigInt(Number.MIN_SAFE_INTEGER)) {
```
fubhy, line 1730: Should we keep internal consts for these around? Is that microoptimization necessary?

```ts
const maxSafeInteger = BigInt(Number.MAX_SAFE_INTEGER)
const minSafeInteger = BigInt(Number.MIN_SAFE_INTEGER)
```

gcanti: @fubhy since it came to your mind, I would say go for it

gcanti (approved):

Issue #380, opened 2023-08-05
S.firstOfArray to extract the first element of a schema to match a given schema


jessekelly881:
I found myself needing to extract the first element of an array from api request that matched a given schema and suspect that something similar might be useful as part of the core lib. Actually, similar utility fns might be useful in general. 

```ts
import * as S from "@effect/schema/Schema";
import * as PR from "@effect/schema/ParseResult";
import * as RA from "@effect/data/ReadonlyArray";

/**
 * Finds the first el in an array that matches the schema.
 */
const firstOfArray = <I, A>(schema: S.Schema<I, A>) =>
  S.transformResult(
    S.array(schema),
    schema,
    (s) => {
      const res = RA.findFirst(s, S.is(schema));
      if (res._tag === "Some") {
        return PR.success(S.encodeSync(schema)(res.value));
      } else return PR.failure(PR.type(schema.ast, s));
    },
    (s) => PR.success([S.decodeSync(schema)(s)])
  );

const s = S.struct({
  descriptions: firstOfArray(
    S.struct({
      lang: S.literal("en"),
      description: S.string,
    })
  ),
});

const res = S.parseSync(s)({
  descriptions: [{ lang: "en", description: "des" }],
});

console.log(res); // { descriptions: { lang: 'en', description: 'des' } }
```

PR #389, opened 2023-08-18
WIP: added S.findAll, S.findFirst, S.findLast

jessekelly881:
Adds S.findAll that finds all values matching a schema A in a schema I[] ignoring cases where the value doesn't match the inner schema. A WIP while awaiting feedback. I think findFirst and findLast fns would be useful as well. 

```ts
 import * as S from "@effect-ts/schema";

 const IntFromString = S.NumberFromString.pipe(S.int())
 const schema = S.findAll(S.array(S.string), IntFromString) // string[] -> int[]

 expect(S.parseSync(schema)(["1", "a", "2", "1.5"])).toEqual([1, 2])
 expect(S.encodeSync(schema)([1])).toEqual(["1"])
```

https://discord.com/channels/795981131316985866/1140667727791542395
https://github.com/Effect-TS/schema/issues/380

### DIFF
--- a/.changeset/curvy-suns-brake.md
+++ b/.changeset/curvy-suns-brake.md
@@ -1,0 +1,5 @@
+---
+"@effect/schema": patch
+---
+
+added S.findAll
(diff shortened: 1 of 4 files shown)


jessekelly881:
@gcanti What is your opinion on these types of transforms? Do you think compiling a decoder within a transform like this is is ok? Obviously this is missing docs, etc. but do you think this is something you would be interested in adding to the repo?

PR #420, opened 2023-09-11
have Schema.Class constructors implement Schema directly

tim-smart:


pigoz:
This is a *huge* DX improvement 👏

PR #430, opened 2023-09-18
require explicit type parameter for dx

fubhy:
@tim-smart It's super easy to forget passing in the self-referencing `S.Class<>` type parameter (./cc @datner)

I'm not sure if there's a better way to solve this dx issue but this at least blows up if you forget to pass it ;-)

### DIFF
--- a/.changeset/lazy-pigs-cross.md
+++ b/.changeset/lazy-pigs-cross.md
@@ -1,0 +1,5 @@
+---
+"@effect/schema": minor
+---
+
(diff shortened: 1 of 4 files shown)


fubhy:
🧸

tim-smart:
👍🏻

Could you do it for extend / transform too?

Then the usual docs and changeset :)

tim-smart:
Tested it out and added error messages. Let me know what you think.

fubhy:
> Tested it out and added error messages. Let me know what you think.

Nice. Good idea!

PR #526, opened 2023-10-30
chore: updated docgen

jessekelly881:


docgen.json
```diff
@@ -1,4 +1,5 @@
 {
+  "$schema": "node_modules/@effect/docgen/schema.json",
   "exclude": ["src/internal/**/*.ts"],
   "theme": "mikearnaldi/just-the-docs",
```
gcanti, line 4: `theme`: while we are at it we can also remove this entry which is the default now

gcanti (changes requested):

gcanti (approved):

gcanti (approved):

PR #533, opened 2023-10-31
expose Equivalence compiler

gcanti:


jessekelly881:
I think having a helper like [this](https://github.com/jessekelly881/effect-schema-compilers/blob/ab41733e1ac4da1668bdcba8140970d58055f574/tests/equivalence.test.ts#L87-L94) would be helpful.

gcanti:
@jessekelly881 https://github.com/Effect-TS/schema/pull/533/files#diff-54dec539aad7e644bf1e312230fb30da27d9ebc7c17e055ae7ec229383d5b8c4R1509

src/Equivalence.ts
```diff
@@ -0,0 +1,196 @@
...
+            if (!tail[j](a[i], b[i])) {
+              return false
+            }
+          }
+        }
+        return true
+      })
+    }
+    case "TypeLiteral": {
+      if (ast.propertySignatures.length === 0 && ast.indexSignatures.length === 0) {
```
kalda341, line 119: What's the purpose of this check? I would have thought that {} should equal {}?

src/Equivalence.ts
```diff
@@ -0,0 +1,196 @@
...
+        // ---------------------------------------------
+        // handle index signatures
+        // ---------------------------------------------
+        for (let i = 0; i < indexSignatures.length; i++) {
+          const is = ast.indexSignatures[i]
+          const aKeys = Internal.getKeysForIndexSignature(a, is.parameter)
+          const bKeys = Internal.getKeysForIndexSignature(b, is.parameter)
+          if (aKeys.length !== bKeys.length) {
+            return false
+          }
```
kalda341, line 152: Wouldn't you also need to check that the keys are equal, not just that their lengths are equal for the following check? Otherwise the resulting indexSignatures[i](a[key], b[key]) call could be passed undefined.

gcanti: yup, forgot a few checks, thanks

src/Equivalence.ts
gcanti: `{}` is quite lax actually

```ts
export const x1: {} = {}
export const x2: {} = { a: 1 }
export const x3: {} = { a: 1, b: () => 1 }
export const x4: {} = [1, 2, 3]
```

PR #535, opened 2023-11-02
feat: added Schema.URL

jessekelly881:
Adds Schema.URL for decoding a string into an instance of URL.

### DIFF
--- a/.changeset/loud-suits-hug.md
+++ b/.changeset/loud-suits-hug.md
@@ -1,0 +1,5 @@
+---
+"@effect/schema": patch
+---
+
+added Schema.URL
--- a/README.md
+++ b/README.md
@@ -2013,6 +2013,28 @@ validate(new Date(0)); // new Date(0)
 validate(new Date("fail")); // throws
 ```
 
+### URL transformations
+
+#### URL
+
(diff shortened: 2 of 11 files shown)


gcanti:
Not sure we want this, after searching "URL" in zod (https://github.com/colinhacks/zod/issues?q=URL) which uses the same technique (`new URL(...)`) I can find the following (non exaustive) list of issues:

- "bug: z.string().url() is incosistent across browser" issue nr. 2256
- "BUG: zod zccepts invalid url: z.string().url().safeParse("a:")" issue nr. 2477
- "The .url validation accepts javascript: schemas which could lead to cross-site scripting vulnerabilities" issue nr. 2353
- "z.string().url() seems to accept any string" issue nr. 2236
- "Support validation of relative URLs" issue nr. 2025
- "Invalid Url when validating a MSSQLServer connection string" issue nr. 2024

jessekelly881:
Hmm. I see your point but a lot of these issues come from a misunderstanding of how `new URL()` works or not knowing that `.url()` is just a simple wrapper around `new URL()` in the first place. What if we just put a warning in the docs?

PR #574, opened 2023-11-19
add TaggedRequest class

tim-smart:


src/Schema.ts
```diff
@@ -3691,6 +3692,77 @@ export const TaggedError = <Self>() =>
   )
 }
 
+function RequestClass(this: any, props: any) {
```
tim-smart, line 3695: Will wait for upstream effect release to replace this with Request.Class

PR #575, opened 2023-11-20
[WIP] add support for `Duration`

fubhy:
@gcanti What kind of constructors and transformations would a user likely expect and use here?

Is it fine to only offer `hrtime` as the standard wire format and always transform that to `nanos` as the internal representation?

Do you see any downsides with that other than optimization (miniscule byte size overhead of the wire format for cases where would technically not be required)

### DIFF
--- a/src/Schema.ts
+++ b/src/Schema.ts
@@ -6,6 +6,7 @@ import * as BigInt_ from "effect/BigInt"
 import * as Brand from "effect/Brand"
 import * as Chunk from "effect/Chunk"
 import * as Data from "effect/Data"
+import * as Duration from "effect/Duration"
 import type * as Effect from "effect/Effect"
 import * as Either from "effect/Either"
 import * as Encoding from "effect/Encoding"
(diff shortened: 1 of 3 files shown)


gcanti:
IMO that's fine as a default encoding

src/Schema.ts
```diff
@@ -2810,6 +2811,88 @@ export const NonNegativeBigint: Schema<string, bigint> = bigint.pipe(
...
+        fc.bigUint().map((_) => Duration.nanos(_)),
+        fc.bigUint().map((_) => Duration.micros(_)),
+        fc.maxSafeNat().map((_) => Duration.millis(_)),
+        fc.maxSafeNat().map((_) => Duration.seconds(_)),
+        fc.maxSafeNat().map((_) => Duration.minutes(_)),
+        fc.maxSafeNat().map((_) => Duration.hours(_)),
+        fc.maxSafeNat().map((_) => Duration.days(_)),
+        fc.maxSafeNat().map((_) => Duration.weeks(_))
+      ),
+    [Internal.EquivalenceHookId]: () => ReadonlyArray.getEquivalence(Equivalence.strict())
```
gcanti, line 2847: Not sure this is ok, tests are missing

gcanti (changes requested):

src/Schema.ts
fubhy: Yep. This was just copy & paste to ask for feedback of the overall approach. Will add tests now and plow ahead.

src/Parser.ts
```diff
@@ -359,7 +359,7 @@ const go = (ast: AST.AST, isDecoding: boolean): Parser<any, any> => {
       }
       return (input: unknown, options) => {
         if (!Array.isArray(input)) {
-          return ParseResult.failure(ParseResult.type(unknownArray, input))
+          return ParseResult.failure(ParseResult.type(ast, input))
```
fubhy, line 362: @gcanti Is this fine?

fubhy, line 362: @gcanti How would you suggest we solve this sustainably? I believe some kind of API is missing here to allow to pass fallback annotations instead of just a message (current third argument).

PR #602, opened 2023-11-24
updated BigDecimal Pretty instance

jessekelly881:
There is an inconsistency with how BigDecimals are rendered to a string. @fubhy Do you think we should wrap the string in `"BigDecimal()"` or just render it to a string directly? I'm leaning towards rendering it directly to a string..

### DIFF
--- a/.changeset/chilly-cobras-jump.md
+++ b/.changeset/chilly-cobras-jump.md
@@ -1,0 +1,5 @@
+---
+"@effect/schema": patch
+---
+
+updated BigDecimal Pretty instance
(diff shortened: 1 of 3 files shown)


fubhy:
I'll defer to @gcanti. I'm not sure if pretty printing is supposed to return a decodable string or if it's supposed to also carry a hint of which data type it represents. My gut tells me that there's value in adding that hint when pretty printing but yeah, defferring to Giulio here.

gcanti:
Yeah I think that hinting at the source data type is pretty useful

jessekelly881:
Ok! Leaving it as it is then.

PR #604, opened 2023-11-24
feat: added Duration filters

jessekelly881:


jessekelly881:
@fubhy Do you mind taking a look at this when you have a sec? :)

README.md
```diff
@@ -2118,6 +2131,75 @@ parse("0"); // BigDecimal(0)
...
+import * as S from "@effect/schema/Schema";
+import * as Duration from "effect/Duration";
+
+// $ExpectType Schema<bigint, Duration>
+const schema = S.DurationFromBigint;
+
+const parse = S.parseSync(schema);
+parse(0n); // 0 seconds
+parse(5000000000n); // 5 seconds
+```
```
fubhy, line 2182: This makes sense but it could also be a massive footgun for ppl not reading docs carefully. The fact that numerical conversion depends on the language primitive (biging vs. number) alone is treated as miliseconds or nanoseconds depending on the used primitive is not super intuitive.

docs/modules/Schema.ts.md
```diff
@@ -710,6 +726,32 @@ export declare const Duration: Schema<readonly [seconds: number, nanos: number],
...
 
+## DurationFromBigint
+
+A schema that transforms a `bigint` tuple into a `Duration`.
+Treats the value as the number of nanoseconds.
+
+**Signature**
+
+```ts
+export declare const DurationFromBigint: Schema<bigint, Duration.Duration>
```
fubhy, line 737: Maybe we should call this `DurationFromBigIntNanos` for the aforementioned reason?

docs/modules/Schema.ts.md
```diff
@@ -710,6 +726,32 @@ export declare const Duration: Schema<readonly [seconds: number, nanos: number],
...
+
+## DurationFromNumber
+
+A schema that transforms a `number` tuple into a `Duration`.
+Treats the value as the number of milliseconds.
+
+**Signature**
+
+```ts
+export declare const DurationFromNumber: Schema<number, Duration.Duration>
```
fubhy, line 750: And `DurationFromNumberMillis` here?

Thoughts @gcanti?

docs/modules/Schema.ts.md
```diff
@@ -720,8 +762,104 @@ export declare const DurationFromSelf: Schema<Duration.Duration, Duration.Durati
...
+
+## durationFromBigint
+
+A combinator that transforms a `bigint` into a `Duration`.
+Treats the value as the number of nanoseconds.
+
+**Signature**
+
+```ts
+export declare const durationFromBigint: <I, A extends bigint>(self: Schema<I, A>) => Schema<I, Duration.Duration>
```
fubhy, line 858: Same here.

docs/modules/Schema.ts.md
```diff
@@ -736,6 +874,19 @@ export declare const durationFromHrTime: <I, A extends readonly [seconds: number
...
 
+## durationFromNumber
+
+A combinator that transforms a `number` into a `Duration`.
+Treats the value as the number of milliseconds.
+
+**Signature**
+
+```ts
+export declare const durationFromNumber: <I, A extends number>(self: Schema<I, A>) => Schema<I, Duration.Duration>
```
fubhy, line 885: Same here.

fubhy (changes requested):
This is all good imho except for that comment on API ambiguity for nanos vs. millis on bigint vs. number.

README.md
jessekelly881: I agree. Although, this is what is currently is done with Duration.decode. I modeled it after that. I'll update it to use fromMillis, etc. I was debating that as well.

dtslint/Schema.ts
```diff
@@ -772,3 +772,25 @@ S.BigDecimalFromNumber
 
 // $ExpectType Schema<string, BigDecimal>
 S.bigDecimalFromString(S.string)
+
+// ---------------------------------------------
+// BigDecimal
```
gcanti, line 777: typo

docs/modules/Schema.ts.md
gcanti: Maybe just `DurationFromMillis` and `DurationFromNanos`

jessekelly881: I like this more. Will update it in a bit.

fubhy (approved):

PR #605, opened 2023-11-25
feat: added toString for Schema classes

jessekelly881:
Adds a default toString method to schema classes that uses the compiled Pretty instance.

### DIFF
--- a/src/Schema.ts
+++ b/src/Schema.ts
@@ -4446,7 +4446,7 @@ const makeClass = <I, A>(
     static [TypeId] = variance
 
     toString() {
-      return pretty(this as any)
+      return `${this.constructor.name}(${pretty(this as any)})`
     }
 
     static pipe() {
--- a/src/Schema.ts
+++ b/src/Schema.ts
@@ -4462,7 +4462,8 @@ const makeClass = <I, A>(
             ParseResult.succeed(input)
             : ParseResult.fail(ParseResult.type(ast, input)), {
           [AST.DescriptionAnnotationId]: `an instance of ${this.name}`,
-          [hooks.PrettyHookId]: () => (props: any) => new this(props).toString(),
+          [hooks.PrettyHookId]: (struct: any) => (self: any) =>
+            `${self.constructor.name}(${struct(self)})`,
           [hooks.ArbitraryHookId]: (struct: any) => (fc: any) =>
             struct(fc).map((props: any) => new this(props))
         }),
--- a/src/Schema.ts
+++ b/src/Schema.ts
@@ -40,7 +40,7 @@ import * as InternalSchema from "./internal/schema.js"
 import * as InternalSerializable from "./internal/serializable.js"
 import * as Parser from "./Parser.js"
 import * as ParseResult from "./ParseResult.js"
-import type { Pretty } from "./Pretty.js"
+import * as Pretty from "./Pretty.js"
 import type * as Serializable from "./Serializable.js"
 
 // ---------------------------------------------
@@ -1503,7 +1503,7 @@ export interface FilterAnnotations<A> extends DocAnnotations<A> {
    */
   readonly jsonSchema?: AST.JSONSchemaAnnotation
   readonly arbitrary?: (...args: ReadonlyArray<Arbitrary<any>>) => Arbitrary<any>
-  readonly pretty?: (...args: ReadonlyArray<Pretty<any>>) => Pretty<any>
+  readonly pretty?: (...args: ReadonlyArray<Pretty.Pretty<any>>) => Pretty.Pretty<any>
   readonly equivalence?: () => Equivalence.Equivalence<A>
 }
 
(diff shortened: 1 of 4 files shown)


tim-smart:
This should be done upstream in the effect/Data module :)

jessekelly881:
@tim-smart Can it be though? Sense it uses the Pretty module?

jessekelly881:
@tim-smart :(

test/Schema/Class.test.ts
```diff
@@ -208,6 +210,8 @@ describe("Schema/Class", () => {
 
   it("TaggedClass", () => {
     let person = new TaggedPersonWithAge({ id: 1, name: "John", age: 30 })
+
+    expect(String(person)).toEqual(`{ "_tag": "TaggedPerson", "id": 1, "age": 30, "name": "John" }`)
```
tim-smart, line 214: I think we need to update the Pretty output to make it feel like a class instance rather than a struct. I.e. the node console.log output is something like `TaggedPerson { id: 1, ... }`

jessekelly881: I like the idea but I don't think you can do this inside makeClass generally since e.g.

```ts
class Person extends S.Class<Person>()({
  id: S.number,
  name: S.string.pipe(S.nonEmpty()),
}) {}
```

doesn't pass a name. We could check for a tag but it might make for somewhat inconsistent results. `Name { ... } ` when using `TaggedClass` and just `{ ... }` when using `Class`.

tim-smart: You can use function.name for Schema.Class?

jessekelly881: Oooh good point. Let me try that. :)

tim-smart: I think `TaggedPerson({ ... })` is a good format, as it will match BigDecimal and Duration.

src/Schema.ts
```diff
@@ -4441,6 +4445,10 @@ const makeClass = <I, A>(
 
     static [TypeId] = variance
 
+    toString() {
+      return `${this.constructor.name}(${pretty(this as any)})`
```
tim-smart, line 4449: Can this be added to the ast as a pretty hook?

tim-smart: I was thinking this:

```diff
diff --git a/.vscode/settings.json b/.vscode/settings.json
index 55a04b7..f33b509 100644
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -22,13 +22,9 @@
   "[typescriptreact]": {
     "editor.defaultFormatter": "dbaeumer.vscode-eslint"
   },
-  "eslint.validate": [
-    "markdown",
-    "javascript",
-    "typescript"
-  ],
+  "eslint.validate": ["markdown", "javascript", "typescript"],
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
   },
   "editor.quickSuggestions": {
     "other": true,
@@ -42,7 +38,7 @@
   "editor.tabCompletion": "off",
   "editor.suggest.localityBonus": true,
   "editor.suggestSelection": "recentlyUsed",
-  "editor.wordBasedSuggestions": true,
+  "editor.wordBasedSuggestions": "matchingDocuments",
   "editor.parameterHints.enabled": true,
   "files.watcherExclude": {
     "**/target": true
diff --git a/src/Schema.ts b/src/Schema.ts
index 3a76808..1c461c5 100644
--- a/src/Schema.ts
+++ b/src/Schema.ts
@@ -4433,7 +4433,6 @@ const makeClass = <I, A>(
   additionalProps?: any
 ): any => {
   const validator = Parser.validateSync(selfSchema)
-  const pretty = Pretty.to(selfSchema)
 
   return class extends Base {
     constructor(props: any = {}, disableValidation = false) {
@@ -4446,7 +4445,7 @@ const makeClass = <I, A>(
     static [TypeId] = variance
 
     toString() {
-      return `${this.constructor.name}(${pretty(this as any)})`
+      return Pretty.to(this.constructor as any)(this)
     }
 
     static pipe() {
@@ -4462,7 +4461,8 @@ const makeClass = <I, A>(
             ParseResult.succeed(input)
             : ParseResult.fail(ParseResult.type(ast, input)), {
           [AST.DescriptionAnnotationId]: `an instance of ${this.name}`,
-          [hooks.PrettyHookId]: () => (props: any) => new this(props).toString(),
+          [hooks.PrettyHookId]: (struct: any) => (self: any) =>
+            `${self.constructor.name}(${struct(self)})`,
           [hooks.ArbitraryHookId]: (struct: any) => (fc: any) =>
             struct(fc).map((props: any) => new this(props))
         }),
```

jessekelly881: Oh yeah that's more inline with other types. const pretty = .. should probably still be defined as before though. I'll update it.

tim-smart:
@gcanti did you have any additional thoughts?

gcanti:
LGTM

PR #617, opened 2023-12-01
add Schema.mapLiterals

matheuspuel:
Something I use often and others may benefit from.
Helps mapping values from external APIs. Similar to an enum.

This one is very common:
```typescript
S.mapLiterals(["F", false], ["T", true])
```


gcanti:
A union of transformations would be better than a transformation between two unions, it retains more info (e.g. the individual mappings `0 <-> 0n`, `"a" <-> true`, `null <-> false` in the example below).

```ts
import * as S from "@effect/schema/Schema"
import * as AST from "@effect/schema/AST"

export const transformLiteral = <From extends AST.LiteralValue, To extends AST.LiteralValue>(
  from: From,
  to: To
): S.Schema<From, To> =>
  S.transform(
    S.literal(from),
    S.literal(to),
    () => to,
    () => from
  )

export const transformLiterals = <
  const A extends ReadonlyArray<readonly [from: AST.LiteralValue, to: AST.LiteralValue]>
>(
  ...pairs: A
): S.Schema<A[number][0], A[number][1]> => S.union(...pairs.map(([from, to]) => transformLiteral(from, to)))

// --------------------

// const schema: S.Schema<0 | "a" | null, bigint | boolean>
const schema = transformLiterals([0, BigInt(0)], ["a", true], [null, false])

const parseSync = S.parseSync(schema)
const encodeSync = S.encodeSync(schema)

console.log(parseSync(0)) // 0n
console.log(encodeSync(0n)) // 0
console.log(parseSync("a")) // true
console.log(encodeSync(true)) // "a"
console.log(parseSync(null)) // false
console.log(encodeSync(false)) // null

console.log("%o", schema.ast)
/*
{
  _tag: 'Union',
  types: [
    {
      _tag: 'Transform',
      from: { _tag: 'Literal', literal: 0, ... },
      to: { _tag: 'Literal', literal: 0n, ... },
      ...
    },
    {
      _tag: 'Transform',
      from: { _tag: 'Literal', literal: 'a', ... },
      to: { _tag: 'Literal', literal: true, ... },
      ...
    },
    {
      _tag: 'Transform',
      from: { _tag: 'Literal', literal: null, ... },
      to: { _tag: 'Literal', literal: false, ... },
      ...
    }
  ],
  ...
}
*/
```

gcanti:
IMO we can also export `transformLiteral`

gcanti:
Thanks @matheuspuel

PR #623, opened 2023-12-04
fix: expose `JsonSchema7Top` and `JsonSchema7` dts

sukovanej:
I started seeing `Cannot find name 'JsonSchema7Top'.`

### DIFF
--- a/.changeset/calm-rules-refuse.md
+++ b/.changeset/calm-rules-refuse.md
@@ -1,0 +1,5 @@
+---
+"@effect/schema": patch
+---
+
+Export `JsonSchema7` types.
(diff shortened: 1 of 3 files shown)


gcanti:
> Do you mean to export and generate docs also for all the JsonSchema7Any, JsonSchema7Unknown, ... parts of the JsonSchema7?

yeah, so it will clarify the subset of JSON schema that we support

> Regarding the JsonSchema7Top - I noticed some JSON Schema documents are talking about a root schema so maybe we could call it a JsonSchema7Root but I'm not sure this is actually the established term.

ok, I'll do it in the next minor release, thanks

gcanti:
Thanks @sukovanej

PR #687, opened 2023-12-25
README: add email example

nikgraf:
Coming from Zod I felt a bit frustrated that some useful string validations are not available. While I would be concerned adding them to the core I was thinking we can at least make it easy by adding useful examples and developers can decide by themselves if this is for them.

Note I took this one because apparently Zod experimented with it and ended up with this one. Related code here: https://github.com/colinhacks/zod/blob/3e4f71e857e75da722bd7e735b6d657a70682df2/src/types.ts#L555-L568

### DIFF
--- a/README.md
+++ b/README.md
@@ -2790,6 +2790,20 @@ parse([
 ]); // new Map([[1, "a"], [2, "b"], [3, "c"]])
 ```
 
+# Useful Examples
+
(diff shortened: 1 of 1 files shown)


fubhy:
Moved to https://github.com/Effect-TS/effect/pull/1844

